Issue #54, opened 2015-10-25
RVIZ plugin -  Display the status of an E-Stop topic in RVIZ 


thaeds:
We need an Rviz plugin which shows a single icon, which displays either an E-Stop detected, or E-Stop not detected image, depending on what is published to the /robot_enabled topic.

Rviz uses the rqt/Qt framework for displaying elements on the screen. It uses message passing with Signals and Slots.

Here is a good tutorial to help get you started with Rviz Plugins.
http://docs.ros.org/jade/api/rviz_plugin_tutorials/html/panel_plugin_tutorial.html

PR #61, opened 2015-11-22
eStop_plugin is added.

indirectlogic:


rviz_plugins/eStop_plugin/src/bare_panel.cpp
```diff
@@ -0,0 +1,45 @@
...
+{
+	QHBoxLayout* topic_layout = new QHBoxLayout;
+	topic_layout->addWidget( new QLabel( "eStop Status:" ) );
+	output_topic_editor_ = new QLabel("TEST");
+	topic_layout->addWidget( output_topic_editor_ );
+
+	QVBoxLayout* layout = new QVBoxLayout;
+	layout->addLayout( topic_layout );
+	setLayout( layout );
+	output_topic_editor_->setText("COOLBEFORE");
```
alchaussee: This line isnt needed

rviz_plugins/eStop_plugin/src/bare_panel.cpp
```diff
@@ -0,0 +1,45 @@
...
+		output_topic_editor_->setText("disabled");
+	}
+}
+
+BarePanel::BarePanel( QWidget* parent )
+  : rviz::Panel( parent )
+{
+	QHBoxLayout* topic_layout = new QHBoxLayout;
+	topic_layout->addWidget( new QLabel( "eStop Status:" ) );
+	output_topic_editor_ = new QLabel("TEST");
```
alchaussee: The QLabel constructor doesn't have to have any arguments.
The label can remain empty until the first callback is called or it could start as "enabled."

rviz_plugins/eStop_plugin/manifest.xml
```diff
@@ -0,0 +1,19 @@
+<package>
+  <description brief="RViz plugin tutorials">
```
alchaussee: Make description and brief relevent

rviz_plugins/eStop_plugin/manifest.xml
```diff
@@ -0,0 +1,19 @@
...
+
+     Tutorials showing how to write plugins for RViz.
+
+  </description>
+  <author>Dave Hershberger</author>
```
alchaussee: change the author to your name

rviz_plugins/eStop_plugin/manifest.xml
```diff
@@ -0,0 +1,19 @@
+<package>
```
alchaussee: This is fine for now but we should change it so there is only one manifest file and one plugin description file for all of the plugins like this https://github.com/ros-visualization/visualization_tutorials/tree/groovy-devel/rviz_plugin_tutorials

rviz_plugins/eStop_plugin/src/bare_panel.h
```diff
@@ -0,0 +1,37 @@
...
+  {
+    Q_OBJECT
+    public:
+      BarePanel( QWidget* parent = 0 );
+
+    private:
+      void subCallback(const std_msgs::Bool &msg);
+
+    public Q_SLOTS:
+
```
alchaussee: I think you can remove the Q_SLOTS lines since you arent using Q_SLOTS

alchaussee:
should probably change the name of the files from bare_panel to something like estop_panel

rviz_plugins/eStop_plugin/.gitignore
```diff
@@ -0,0 +1,3 @@
+/bin/
+/build/
+/lib/
```
barulicm: @alchaussee We should figure out how to merge these into the main .gitignore. Not a huge fan of cascading gitignores if it can be avoided.

rviz_plugins/eStop_plugin/plugin_description.xml
```diff
@@ -0,0 +1,9 @@
+<library path="lib/libeStop_plugin">
+  <class name="bare/eStop"
+         type="bare::BarePanel"
+         base_class_type="rviz::Panel">
+    <description>
+      A panel widget that displats the eStop status.
```
barulicm: spelling error "displats"

rviz_plugins/eStop_plugin/src/bare_panel.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+	sub = nh_.subscribe("/robot_enabled", 1, &BarePanel::subCallback, this);
+
+	//connect( this, SIGNAL( changeText() ), output_topic_editor_, SLOT( setTextLabel() ));
+}
+
+}
+
+#include <pluginlib/class_list_macros.h>
+PLUGINLIB_EXPORT_CLASS(bare::BarePanel,rviz::Panel )
```
barulicm: Move these two lines above the namespace declaration to match other export macros in the code. The #include line should go with the other #include lines, and the macro should go right above the namespace keyword.

rviz_plugins/eStop_plugin/src/bare_panel.h
```diff
@@ -0,0 +1,37 @@
+#ifndef BARE_PANEL_H
+#define BARE_PANEL_H
+
+#include <ros/ros.h>
+#include <QLabel>
+#include <std_msgs/Bool.h>
+
+#include <rviz/panel.h>
+
+class QLineEdit;
```
barulicm: Unused type QLineEdit

rviz_plugins/eStop_plugin/src/bare_panel.cpp
```diff
@@ -0,0 +1,45 @@
+#include <stdio.h>
+#include <string>
+
+#include <QVBoxLayout>
+#include <QHBoxLayout>
+#include <QLineEdit>
```
barulicm: Unused include QLineEdit

rviz_plugins/eStop_plugin/src/bare_panel.h
```diff
@@ -0,0 +1,37 @@
...
+      void subCallback(const std_msgs::Bool &msg);
+
+    public Q_SLOTS:
+
+    protected Q_SLOTS:
+
+    protected:
+      ros::Subscriber sub;
+      ros::NodeHandle nh_;
+      QLabel* output_topic_editor_;
```
barulicm: It looks like this may have been an editable field in a previous version, but the current variable name does not reflect the current function of the member. Consider a different name.

rviz_plugins/eStop_plugin/src/bare_panel.h
```diff
@@ -0,0 +1,37 @@
...
+    private:
+      void subCallback(const std_msgs::Bool &msg);
+
+    public Q_SLOTS:
+
+    protected Q_SLOTS:
+
+    protected:
+      ros::Subscriber sub;
+      ros::NodeHandle nh_;
```
barulicm: There are two conventions for private or protected members in the code right now are either a single underscore preceding the member name or no extra punctuation added to the member name. (ie. _nh or nh).
Please change these names to match one of these conventions and make sure you only use one for the whole file.

rviz_plugins/eStop_plugin/src/bare_panel.h
rmkeezer: Q_SLOTS are actually necessary for this because the qt GUI objects communicate with Q_SLOTS.

rmkeezer:
closes #54

PR #77, opened 2015-12-01
Add CI To igvc-software

### LINKED ISSUE #40: Fix build dependency orders related to igvc_msgs
On first build, some nodes in IGVC get built before igvc_msgs finishes, causing them to error out. Currently, all builds after the first one work fine, but we should still probably fix this.

jgkamat:
This is almost complete, but a few more things need to be done before this will work. 
Right now, the tests fail only when the code dosen't compile. (or when dependencies change...)

The flycapture stuff is very fragle, and is very likely to break if anything moves around/changes dependencies. This can all be changed through `indigo-setup.sh` though. (by the way, indigo-setup should work to install all your dependencies on a clean ros install).

- [x] Add @georgeburdell to writable members on this repo so he can update our status tokens
- [x] Add tests (is there a `catkin_make tests` for igvc?)
- [x] Find which files dependencies are stored in (which files do you update if you want to change what `rosdep install igvc` does?)
- [x] Do you guys want any form of deployment? (Auto-updating docs pages, etc)
- ~~Maybe Fix #40...~~

This is all pretty new so there's a chance there could be bugs. I'll fix them as they come up. Let me know what you think!


util/indigo-setup.sh
```diff
@@ -0,0 +1,27 @@
...
+wget https://files.gitter.im/RoboJackets/igvc-software/BGRW/flycapture2-2.8.3.1-amd64-pkg.tgz
+tar -xvf flycapture2-2.8.3.1-amd64-pkg.tgz
+cp ./src/util/flycapture-setup.exp flycapture2-2.8.3.1-amd64/
+cp ./src/util/flycapture-setup2.exp flycapture2-2.8.3.1-amd64/
+cd flycapture2-2.8.3.1-amd64
+
+sudo ./flycapture-setup.exp || true		# Cross our fingers...
+sudo apt-get install -yf
+sudo ./flycapture-setup.exp || true		# Cross our fingers...
+sudo apt-get install -yf
```
alchaussee: why do these commands get called twice

alchaussee:
I feel like the submodule should be a Robojackets repo not your own

util/indigo-setup.sh
jgkamat: This is because the flycapture setup is bugged, and needs to be run 3 times before it works...
There might(?) be a command that automatically resolves dependencies, but since I'm running their script I can't change it to that.

jgkamat:
The submodule has been generalized to the point where it can be used with any repository so I didn't think it would make sense to go in the RJ namespace. I can't move it now, because that would break a bunch of builds (I'm using it in >5 of my own projects at the very least).

If you want, we could fork it to the RJ namespace or create a gitlab RoboJackets namespace, and create a repo mirror (https://about.gitlab.com/2015/11/22/gitlab-8-2-released/).

util/indigo-setup.sh
jgkamat: If there's a better way to run this, let me know and I'll update this script.

jgkamat:
This should be good now. The builds take a LOOOONG time though.

Issue #58, opened 2015-10-25
Create an node to transform rectified camera images into overhead images 


thaeds:
Create this node in the bumblebee2 folder. You need to subscribe to /left/image_rect_color, and publish to /image_overhead_color.

Here is a tutorial about using compute homography to turn an image from one perspective into an image from another perspective.
http://docs.opencv.org/doc/tutorials/features2d/feature_homography/feature_homography.html

Issue #36, opened 2015-04-11
Write a ROS node to detect Simulated Potholes


barulicm:
Rules sec. II.3 Obstacle Course:
"Simulated potholes of 2 foot diameter solid white circles may be inserted. These simulated pot must be avoided or an end of run will occur."

PR #91, opened 2016-02-28
Pothole Detector

nareddyt:
Implemented the Pothole Detector. Also includes an overhead transform in the bumblebee node, which is not currently used.


nareddyt:
Closes issues #58 and #36

PR #99, opened 2016-05-16
Upstream

dpattison3:
This adds a localization node that takes in all of our positional data to give one filtered version. Has not been tested for efficacy, but does run without errors. (It can't be tested on any of our rosbag data because it has conventions that I needed to change)


thaeds:
great work A+++++ would merge again

Issue #104, opened 2016-08-08
Reorganize launch files


jzheng84:
Simplify launch procedure so that everyone can easily launch the robot

JasonGibson274:
There should be a launch file that will launch everything needed to drive the robot forward (motor controller, light controller, etc) I should be able to publish to /motors and the robot will drive.

JasonGibson274:
add a motor controller node to the joystick launch file so that all that needs to be launched to joystick robot around is one file

JasonGibson274:
ensure that all sensors are launched in sensors.launch

JasonGibson274:
All of the specific things mentioned are solved and any specific changes should be their own issue

PR #115, opened 2016-09-17
modified odometry node to publish a 0,0,0 position to be consistent w…

dpattison3:
…ith the frame we are publishing - added covariance matrix stand in - TODO update covariance matrix with true values


dpattison3:
This is a small set of changes, but to be able to get data to do covariance correctly these changes must be on the robot.

rmkeezer:
Related to #104

PR #122, opened 2016-10-18
Stop building igvc_msgs before all

jgkamat:
See #40 

Because that was fixed, you guys don't need a hack build target.


rmkeezer:
Good thinking

PR #124, opened 2016-10-20
SearchProblem Update

red-eight:
Uses construction initialization list for Path constructor with arguments.
Throw exception classes instead of char*.
Avoid undefined behavior in getLastState() by making sure `states` is not empty.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@
 /sandbox/ardupilot/config.mk
 
 *~
+*.swp
 
 #Qt Creator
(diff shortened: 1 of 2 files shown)


red-eight:
Hello,
Just an important heads up, if there are any catch statements that are catching `const char*` or `char*`, they will miss these `std::out_of_range` exceptions. I recommend removing all statements that throw `char*` and replacing all catch statements with a `catch(const std::exception &e)`.

PR #126, opened 2016-10-20
changed location of the camera calibration files

dpattison3:
not super important, just figured we can put these files back to where the old calibration files were - especially considering I didn't realize we had them (looking in wrong place)


rmkeezer:
Works for me

PR #128, opened 2016-10-23
Fix CI Passing when it should have failed

jgkamat:
Also upgraded DoCIF

@rmkeezer you can merge this when you want :P


rmkeezer:
Nice! The X's are back! I missed them

Issue #132, opened 2016-10-26
Generic Computer Vision Module


rmkeezer:
Create a module that contains common methods used by computer vision modules (lines, potholes, flags)

This may include blocking out parts of the robot from the camera image, blocking out the barrels, converting 2d image to 3d overhead view, etc.

dpattison3:
Possible repeat of #76 
Regardless, should also consider: #51, #72, #105, and of course #76

rmkeezer:
#136 and #138 solves this so far

PR #136, opened 2016-11-06
Cleaned up Linedetector and created CV util file

rmkeezer:
This might not be the best way to make a utils file, this should be looked into more.


rmkeezer:
Possible solution to #132 but should be improved

PR #137, opened 2016-11-10
Fix pothole detector and update to /usb_cam_center topic

jmschonfeld:
This commit fixes the pothole detector and updates the topic it listens to to "/usb_cam_center". The pothole detector will now publish to "/usb_cam_center/pothole_filt_img", "/usb_cam_center/pothole_thres", and "/usb_cam_center/pothole_cloud"


rmkeezer:
Good job! This was needed for the new cameras, although, it might not work with the old bag file now.

rmkeezer:
Also the failure to compile is unrelated to this see #139

PR #141, opened 2016-11-15
missing dependency 

JasonGibson274:
the ros dependency image_geometry is missing from our deps but it is used in both line detector and pot hole detector


barulicm:
Your first pass was actually correct. Dependencies listed in the CMakeLists file should also appear in the package manifest (package.xml).
This allows tools like rosdep, which work based on the package manifest, to correctly install the dependencies for our packages.

PR #142, opened 2016-11-17
Consolidate toPointCloud methods

jmschonfeld:
Creates MatToContour and toPointCloud methods in CVUtils to consolidate these functions into one file.


rmkeezer:
I approve

Issue #35, opened 2015-04-05
tf Transform on IMU data yielding unexpected results


barulicm:
Applying the tf transform /imu -> /base_footprint on the IMU orientation quaternion does not produce the expected output. Grabbing Euler angles from the quaternion before and after the transformation shows the correct transformation in yaw, but leaves roll and pitch untouched.

barulicm:
Hack-ish fix added to Ardupilot node. See igvc/src/ardupilot/ardupilot.cpp:51

dpattison3:
I haven't been able to recreate this issue with the new IMU set up, so I'm going to close the issue.

Issue #68, opened 2015-11-30
Reduce GPS Noise through Kalman Filter


thaeds:
Another type of filter is acceptable if more appropriate, but GPS noise must be reduced through sensor fusion, possibly through using a Kalman Filter. The problem is observable as either the robot believing it is jumping around its physical location while it is actually stationary (noise), or the robot believing it is moving in a certain direction while it is actually stationary (drift).

Implementing sensor fusion through a Kalman Filter will allow us to estimate our current position with more accuracy and less noise, given that we will be able to take into account our current wheel odometer y.

AlexTrimm:
There should be code sitting around somewhere for this. We attempted it 2-3
years ago, but we were never able to tune it well enough to make it useful.
Matt and Al should be passingly familiar with it. I'd suggest talking to a
controls professor about effective methods for tuning the parameters. Dr.
Feron out of the AE school has been involved in student competition teams
in the past. He might be worth asking if you don't know of someone
personally. If you're looking to learn more about Kalman Filters, the
original paper is quite good.

thaeds:
That sounds awesome - I think the first place to look is the ROS implementation
http://wiki.ros.org/robot_pose_ekf

But @barulicm has noted that in the past this implementation has not been good enough to use.

Certainly reading the original paper would be awesome regardless. (below)
https://www.cs.unc.edu/~welch/kalman/media/pdf/Kalman1960.pdf

barulicm:
Since this issue is coming back, I'll note that it's worth looking into [robot_pose_ekf](http://wiki.ros.org/robot_pose_ekf) again, as a significant amount of development has been put into it since I last looked at it.

jgkamat:
@rmkeezer I think this issue can be closed? Referenced in #144

dpattison3:
It was a little hasty to say that it solved that issue. It addresses the issue, but it still needs to be tested in competition scenarios, so I think it should remain open until I can show it working well in those conditions.

dpattison3:
The localization node has been tested, so I think we can close this now

PR #144, opened 2016-11-22
localization fixes and updates

dpattison3:
improves odometry node
corrects localization launch file
adds imu fixes

The odometry node will still need to be tuned in terms of covariance for new robot, but current values should work. The encoders are on backwards, so currently will produce bad results.
The process noise also needs tuning for the localization node, but needs the GPS to be working (the cable is currently broken).


rmkeezer:
Fixes #35 and #68

PR #146, opened 2016-12-01
Added QUICK WORKAROUND to make rviz_plugins compile

rmkeezer:
A better solution should be found later


jgkamat:
Builds are failing on this branch due to a missing dependency on qt4:

https://192-16996386-gh.circle-artifacts.com/0/tmp/circle-artifacts.kdlLQB9/compile.txt

rmkeezer:
I changed the rvizplugins to use Qt5 instead of Qt4 but the error is still here

jgkamat:
https://196-16996386-gh.circle-artifacts.com/0/tmp/circle-artifacts.PdMRjpN/compile.txt

Seems like it's not finding `igvc_msgs/velocity_pair.h`?

Can you see if thats referenced anywhere?

rmkeezer:
Yes it is needed, but it is also referenced by the motor_controller node and that is building fine

jgkamat:
Your repo is a bit behind master, maybe you could merge master into your repo and see if that helps?

jgkamat:
I just reran the build and it passed, if you have any idea why the previous one failed let me know. They should have been exactly the same.

rmkeezer:
igvc_msgs might have built before igvc, but this race condition should've already been fixed

PR #148, opened 2016-12-04
Update DoCIF to fix caching issues

jgkamat:
There was a regression in the Docker Hub API recently that prevented the code that checks if a cache image is available. This fixes it, and should make your builds faster again.


rmkeezer:
faster builds means faster checkmarks, nice job!

PR #163, opened 2017-01-23
Log dmesg to try to catch compiler failures

jgkamat:
Hopefully this debug output should help me debug your c++ compiler failures.

Let me know if you think anything is wrong with it.


jgkamat:
Also another thing that could be causing your compiler failures is zombie processes, but I don't think that's the issue.

Let me know when you find a failure with this branch merged in and I'll take a peek.

jgkamat:
@rmkeezer have you seen any issues of this compiler error occuring after this was merged in? I also switched the build infrastructure to be backed by 14.04 rather than the old 12.04, so that might have improved things as well.

PR #167, opened 2017-02-02
Adjust and fix pothole detector

jmschonfeld:


rmkeezer:
Nice! This was definitely needed

Issue #149, opened 2016-12-05
Investigate possible race condition in building igvc_msgs


rmkeezer:
CircleCI is sometimes not building igvc_msgs before igvc. This should have been fixed a while ago. If anyone is failing builds on master please post the log here

dpattison3:
Rviz plugings are the only module that doesn't set messages as a dependency (it wasn't compiling when I fixed the earlier issue anyway). Should it depend on messages? That's the only recent thing to change right?

JasonGibson274:
I think we should close this issue and reopen it if the issue returns. CircleCI seems to be building fine to me.

jgkamat:
This is still happening, just at a much reduced rate. [See this build log for details](https://271-16996386-gh.circle-artifacts.com/0/tmp/circle-artifacts.rnZw4Il/compile.txt). I /think/ that rviz_plugins needs to depend on igvc_msgs as well (although, since igvc depends on it, shouldn't igvc_msgs depend on it as well?).

PR #171, opened 2017-02-12
Move Tuning Parameters to YAML File

jmschonfeld:
This moves all of the tuning parameters for the line detector and pothole detector algorithms to two external YAML files. One YAML file is tuned for the gazebo simulator in ideal situations. The other YAML file will be tuned to a real-life scenario in non-ideal conditions. Which YAML file to use is specified in the launch file for line detector and pothole detector.


jgkamat:
@rmkeezer is this failure the igvc_msgs race condition, or just a mistake in the PR? If it's the race condition we should add the hack to build igvc_msgs back (or fix the race condition).

[see the output of this build log](https://271-16996386-gh.circle-artifacts.com/0/tmp/circle-artifacts.rnZw4Il/compile.txt)

jmschonfeld:
The code compiles fine on my VM with no issues, and executes correctly in RViz and Gazebo. The build log fails on files related to the rviz_plugins directory (it looks like), and I know that none of the code I pushed was related to rviz/rviz plugins.

jgkamat:
Yah, this is the race condition we saw a long time ago. It seems like #149 was never fully fixed (just migitated).

jgkamat:
I've rebuilt your build so you get a :white_check_mark:, hopefully we'll fix this issue before anyone sees it agian.

dpattison3:
#149 Fully fixed it at the time, but at the time, we were not building any of the RViz plugins (they didn't compile for other reasons). #173 is the correct solution.
@rmkeezer Be more careful with the code next time.

rmkeezer:
Ah, yeah I added that a long time ago, my bad

rmkeezer:
This is the first step to eliminating magic numbers from our code, good job!

PR #173, opened 2017-02-13
Try to fix igvc_msgs race condition again

### LINKED ISSUE #149: Investigate possible race condition in building igvc_msgs

jgkamat:
Closes #149 

Let me know if this dosen't seem right. No way to be sure if this is fixed without letting it sit for a long time.


rmkeezer:
Ah that was my bad, thanks!

PR #189, opened 2017-03-09
Fix Pothole Detector Transformations

jmschonfeld:
Makes the correct transformations for the pothole detector and removes the hardcoded "optical_cam_center" so that "optical_cam_left", etc. can be used


rmkeezer:
Looks good!

PR #190, opened 2017-03-09
Condensed camera launch files into a single launch file

kscharm:


rmkeezer:
This is related to #104

JasonGibson274:
Can this be changed to follow our conventions on naming of the cameras

Issue #185, opened 2017-03-07
Remove Non-Info Camera From Line Detector


jmschonfeld:
The line detector currently has a hard-coded boolean value that will determine whether to use a camera with info or a camera without info. Since all of our cameras on the robot and in Gazebo currently publish camera info, the code that handles non-info cameras in the line detector should be removed.

PR #191, opened 2017-03-09
Fixes pothole detector seg fault and Removes non-info camera from line detector

jmschonfeld:
Fixes an invalid memory reference (SIGSEGV) caused by accessing a pixel out of the bounds of the image when processing potholes


jmschonfeld:
Fixes issue #185

Issue #188, opened 2017-03-09
Decrease Image Resolution/Frequency


jmschonfeld:
Either decrease the resolution of the image or the frequency of the image processing to ensure that the image processing can run fast enough on the NUCs.

PR #194, opened 2017-03-12
Added spin rate to limit vision modules

rmkeezer:
Slows line detector and pothole detector refresh rate to 5 per second.


rmkeezer:
Fixes #188

Issue #70, opened 2015-11-30
Test and Tune Pothole Detection


thaeds:
The pothole detector is currently tuned for detecting potholes of a certain range of sizes in one particular image. It needs to be tested against many different images and on moving data from multiple data sets, and calibrated/tweeked such that it performs well on all of them.

rmkeezer:
This is a vague issue that has no real goal.
Testing with Gazebo is currently sufficient until the new robot is ready.

rmkeezer:
Although this is vague, pothole detector is detecting circles when just a noise is generated in gazebo grass. It should be more robust than this.

dpattison3:
addressed by #264

PR #195, opened 2017-03-12
Fix Pothole Tuning

jmschonfeld:
The pothole detector will no longer add lines to its point cloud.


rmkeezer:
I reopened the old test and tune pothole detector issue because it needs to be more robust. #70

jmschonfeld:
This updates the pothole detector to filter out much more noise (extra grass textures)

Issue #46, opened 2015-10-18
Detect blue and red flags in images


thaeds:
Towards the end of the advanced course there are red and blue flags that we have to drive between. The lidar is unable to detect the flags so the camera will need to be able to find them.

JasonGibson274:
Cannot use color alone since the color of the barrels too closely matches the color of the red flags

atuttle7:
No longer in the rules. RIP

PR #196, opened 2017-03-12
Added flag detector code. Need to test its current status.

jzheng84:
This was written for the old vehicle.


rmkeezer:
This will be merged so we can then test and make it actually detect flags as per #46

PR #229, opened 2017-08-28
Compeition branch

JasonGibson274:
Open questions: why the gps conversion was removed

Otherwise this is mostly fixing or removing bad code that I wrote at competition/before that should never have been merged into master. Also updating the urdf for Woodi sensor position wise. If we really want to update the urdf to mimic the actual robot then add an issue and assign it to me but I would rather walk someone else (freshmen) through setting up the new robot.


JasonGibson274:
It appears that the UTM conversion node does not support kinetic and the only method we needed was copied into a file. Therefore we do not need this.

dpattison3:
We also have the nav sat transform node that launches with the localization node that performs a similar thing I think.

JasonGibson274:
The header file is only used to convert the latitude and longitude of the waypoint into UTM.

JasonGibson274:
Agreed I will double check the sensors are being launched in a logical fashion and work as intended and I will remove the waypoints file and make a single one.

JasonGibson274:
@rmkeezer we did not have the side cameras attached and we are not really planning on it for the new robot. If we add them in we can revert back to the default being all three.

sandbox/qual_waypoints.csv
```diff
@@ -0,0 +1,2 @@
+42.67833241, -83.19551079
```
JasonGibson274: remove excess waypoints files.

sandbox/test_waypoints.csv
```diff
@@ -0,0 +1,2 @@
+42.67858287, -83.19530227
```
JasonGibson274: remove excess waypoints files.

igvc/launch/sensors.launch
```diff
@@ -6,22 +6,25 @@
 <launch>
```
JasonGibson274: Check that launch file work as intended.

PR #248, opened 2017-10-26
Global mapping

jzheng84:
Probabilistic mapping code. Should be identical to previous code if you set probability threshold to "0". This code only adds points to the global map if we see them for some percentage of previous frames.


igvc/src/global_mapper/main.cpp
```diff
@@ -10,57 +10,90 @@
 #include <ros/ros.h>
 #include <stdlib.h>
 #include <string>
+#include <queue>
```
JasonGibson274: Is this used? It looks like no

igvc/src/global_mapper/main.cpp
```diff
@@ -10,57 +10,90 @@
...
     {
-      pcl::PointXYZ searchPoint(Final[i].x, Final[i].y, Final[i].z);
-      std::vector<int> indices;
-      std::vector<float> distances;
-      if (octree.radiusSearch(searchPoint, searchRadius, indices, distances, 1) == 0)
+      for (auto point : (*pointcloud_list[i]))
       {
-        octree.addPointToCloud(searchPoint, global_map);
-        new_points.push_back(searchPoint);
+        //G and B represent the probability, they are the same value
```
JasonGibson274: Why is the value duplicated? If it is for coloring ignore me

igvc/src/global_mapper/main.cpp
```diff
@@ -10,57 +10,90 @@
...
+        {
+          //Increase probability of the point in the voxel if we see it again
+          int point_idx;
+          float dist;
+          local_octree.approxNearestSearch(point, point_idx, dist);
+          //If the probability is less than the threshold, increase the probability
+          if (local_pc->points[point_idx].g < probability_thresh * 255.0) {
+            local_pc->points[point_idx].g = local_pc->points[point_idx].g + (int)(255.0 / queue_window_size);
+            local_pc->points[point_idx].b = local_pc->points[point_idx].b + (int)(255.0 / queue_window_size);
+          } else if (!firstFrame && local_pc->points[point_idx].g >= probability_thresh * 255.0)
```
JasonGibson274: Can't points be added to the global map multiple times?

igvc/src/global_mapper/main.cpp
```diff
@@ -10,57 +10,90 @@
...
+        //G and B represent the probability, they are the same value
+        //Add a point to the local octree if we haven't seen it before
+        if (!local_octree.isVoxelOccupiedAtPoint(point))
+        {
+          point.r = 255;
+          point.g = 0;
+          point.b = 0;
+          local_octree.addPointToCloud(point, local_pc);
+        } else
+        {
```
JasonGibson274: Is this inline with the ROS guidelines?

igvc/src/global_mapper/main.cpp
```diff
@@ -10,57 +10,90 @@
...
+          float dist;
+          local_octree.approxNearestSearch(point, point_idx, dist);
+          //If the probability is less than the threshold, increase the probability
+          if (local_pc->points[point_idx].g < probability_thresh * 255.0) {
+            local_pc->points[point_idx].g = local_pc->points[point_idx].g + (int)(255.0 / queue_window_size);
+            local_pc->points[point_idx].b = local_pc->points[point_idx].b + (int)(255.0 / queue_window_size);
+          } else if (!firstFrame && local_pc->points[point_idx].g >= probability_thresh * 255.0)
+          {
+            //Add the point to the global map if its probability is greater than or equal to the threshold
+            octree->addPointToCloud(local_pc->points[point_idx], global_map);
```
JasonGibson274: Is this pass by value or by reference. Is this a shared pointer? Because local_pc is destructed when its time in the window expires.

igvc/src/global_mapper/main.cpp
```diff
@@ -10,57 +10,90 @@
...
   }
-  else if (!input->points.empty())
-  {
-    octree.setInputCloud(global_map);
-    pcl::PointCloud<pcl::PointXYZ> initial = *input;
+
+  if (firstFrame && !input->points.empty())
+  { 
+    octree->setInputCloud(global_map);
+    pcl::PointCloud<pcl::PointXYZRGB> initial = *input;
```
JasonGibson274: Why not just use the pointer the entire time?

igvc/src/global_mapper/main.cpp
```diff
@@ -70,7 +103,7 @@ void icp_transform(pcl::PointCloud<pcl::PointXYZ>::Ptr input)
 void frame_callback(const pcl::PointCloud<pcl::PointXYZ>::ConstPtr &msg, const std::string &topic)
 {
   pcl::PointCloud<pcl::PointXYZ>::Ptr transformed =
-      pcl::PointCloud<pcl::PointXYZ>::Ptr(new pcl::PointCloud<pcl::PointXYZ>());
+    pcl::PointCloud<pcl::PointXYZ>::Ptr(new pcl::PointCloud<pcl::PointXYZ>());
   tf::StampedTransform transform;
   if (frames_seen.find(msg->header.frame_id) == frames_seen.end())
   {
```
JasonGibson274: On line 111 two lines down. The transform happens even when it fails to find the transform. I feel like it it fails to find it the method should be over

igvc/src/global_mapper/main.cpp
```diff
@@ -126,8 +163,9 @@ int main(int argc, char **argv)
 
   global_map->header.frame_id = "/odom";
 
-  _pointcloud_pub = nh.advertise<pcl::PointCloud<pcl::PointXYZ>>("/map", 1);
-  _pointcloud_incremental_pub = nh.advertise<pcl::PointCloud<pcl::PointXYZ>>("/map/incremental", 1);
+  _pointcloud_pub = nh.advertise<pcl::PointCloud<pcl::PointXYZRGB>>("/map", 1);
+  _pointcloud_incremental_pub = nh.advertise<pcl::PointCloud<pcl::PointXYZRGB>>("/map/incremental", 1);
+  _pointcloud_prob_pub = nh.advertise<pcl::PointCloud<pcl::PointXYZRGB>>("/prob", 1);
```
JasonGibson274: Is this the same point cloud as above?

igvc/src/line_detector/linedetector.cpp
```diff
@@ -22,15 +22,21 @@ void LineDetector::img_callback(const cv::Mat msg_img, const sensor_msgs::ImageC
...
+
   // Convert image to grayscale
-  cv::cvtColor(src_img, src_img, CV_BGR2GRAY);
+  // cv::cvtColor(src_img, src_img, CV_BGR2GRAY);
 
   // Crops the image (removes sky)
   int topCrop = 2 * src_img.rows / 3;
-  cv::Rect roiNoSky(0, topCrop, src_img.cols, src_img.rows - topCrop);
-  src_img = src_img(roiNoSky);
+  // cv::Rect roiNoSky(0, topCrop, src_img.cols, src_img.rows - topCrop);
```
JasonGibson274: Why did we stop cropping out the sky?

igvc/src/global_mapper/main.cpp
jzheng84: Just for coloring. Initialized as red, and become white as probability increases

igvc/src/global_mapper/main.cpp
jzheng84: Do you mean brackets for conditionals? I'll check

jzheng84: the bracket is consistent with ROS CppStyleGuide, if that is what you mean.

igvc/src/line_detector/linedetector.cpp
jzheng84: @rmkeezer

igvc/src/global_mapper/main.cpp
JasonGibson274: That is weird but okay

igvc/src/global_mapper/main.cpp
```diff
@@ -11,56 +11,93 @@
...
 int maxIter;
-pcl::octree::OctreePointCloudSearch<pcl::PointXYZ> octree(.01f);
+pcl::octree::OctreePointCloudSearch<pcl::PointXYZRGB>::Ptr octree;
 double searchRadius;
+double octree_resolution;
+double probability_thresh;
 
-void icp_transform(pcl::PointCloud<pcl::PointXYZ>::Ptr input)
+//Probabilistic mapping
+//std::vector<pcl::PointCloud<pcl::PointXYZRGB>::Ptr> pointcloud_list;
```
JasonGibson274: can we take this out since it is only a comment?

igvc/src/global_mapper/main.cpp
```diff
@@ -11,56 +11,93 @@
...
+        {
+          //Increase probability of the point in the voxel if we see it again
+          int point_idx;
+          float dist;
+          local_octree.approxNearestSearch(point, point_idx, dist);
+          //If the probability is less than the threshold, increase the probability
+          if (local_pc->points[point_idx].g < probability_thresh * 255.0) {
+            local_pc->points[point_idx].g = local_pc->points[point_idx].g + (int)(255.0 / queue_window_size);
+            local_pc->points[point_idx].b = local_pc->points[point_idx].b + (int)(255.0 / queue_window_size);
+          } else if (!firstFrame && local_pc->points[point_idx].g >= probability_thresh * 255.0
```
JasonGibson274: I am unsure here. Do you want the point only if it exceeds the threshold or if it reaches it. For example if the threshold was 0.6 and the point showed up 3 times then it would not be added, it would have to show up 4 times due to the if else structure. If that is the intention ignore me

igvc/src/global_mapper/main.cpp
```diff
@@ -11,56 +11,93 @@
...
-    icp.align(Final);
-    for (unsigned int i = 0; i < Final.size(); ++i)
+    //Probabilistic calculations here
+    pcl::PointCloud<pcl::PointXYZRGB>::Ptr local_pc = pcl::PointCloud<pcl::PointXYZRGB>::Ptr(new pcl::PointCloud<pcl::PointXYZRGB>());
+    pcl::octree::OctreePointCloudSearch<pcl::PointXYZRGB> local_octree(octree_resolution);
+
+    local_octree.setInputCloud(local_pc);
+
+    std::list<pcl::PointCloud<pcl::PointXYZRGB>::Ptr>::iterator it = pointcloud_list.begin();
+    for (unsigned int i = 0; i < pointcloud_list.size(); i++)
```
JasonGibson274: Since you are already using an iterator and never using i, I would recommend.

`for(auto it = pointcloud_list.begin(); it != pointcloud_list.end(); it++)`

igvc/src/global_mapper/main.cpp
```diff
@@ -11,56 +11,93 @@
...
+      it++;
     }
-    new_points.header.frame_id = "/odom";
-    _pointcloud_incremental_pub.publish(new_points);
+
+    (*local_pc).header.frame_id = "/odom";
+    _pointcloud_prob_pub.publish(*local_pc);
+
+    //Remove first pointcloud
+    //pointcloud_list.erase(pointcloud_list.begin());
```
JasonGibson274: remove commented out line

igvc/src/global_mapper/main.cpp
```diff
@@ -11,56 +11,93 @@
...
+    local_octree.setInputCloud(local_pc);
+
+    std::list<pcl::PointCloud<pcl::PointXYZRGB>::Ptr>::iterator it = pointcloud_list.begin();
+    for (unsigned int i = 0; i < pointcloud_list.size(); i++)
     {
-      pcl::PointXYZ searchPoint(Final[i].x, Final[i].y, Final[i].z);
-      std::vector<int> indices;
-      std::vector<float> distances;
-      if (octree.radiusSearch(searchPoint, searchRadius, indices, distances, 1) == 0)
+      //for (auto point : (*pointcloud_list[i]))
```
JasonGibson274: remove commented out line

igvc/src/global_mapper/main.cpp
```diff
@@ -70,51 +107,53 @@ void icp_transform(pcl::PointCloud<pcl::PointXYZ>::Ptr input)
...
-  }
-  tf_listener->lookupTransform("/odom", msg->header.frame_id, ros::Time(0), transform);
-  pcl_ros::transformPointCloud(*msg, *transformed, transform);
-  for (auto point : *transformed)
-  {
-    point.z = 0;
-  }
+    tf_listener->lookupTransform("/odom", msg->header.frame_id, ros::Time(0), transform);
+    pcl_ros::transformPointCloud(*msg, *transformed, transform);
+    for (auto point : *transformed)
```
JasonGibson274: Why are you doing this? We ignore z in the path planner anyway

JasonGibson274:
@jzheng84 I thought you were leaving out the line detector changes

jzheng84:
@JasonGibson274 I realized that Matt commented out the code that crops AND the code that refills the sky area with black. I uncommented one but forgot to uncomment the other. Now this should be back to how the line detector originally was, and we can fix the lack of robustness at another time.

igvc/src/global_mapper/main.cpp
```diff
@@ -11,56 +11,89 @@
...
+int queue_window_size;
+
+void icp_transform(pcl::PointCloud<pcl::PointXYZRGB>::Ptr input)
 {
-  if (!firstFrame)
+  if (pointcloud_list.size() < queue_window_size)
+  {
+    pointcloud_list.push_back(input);
+  }
+  if (pointcloud_list.size() == queue_window_size)
```
JasonGibson274: comparison between signed and unsigned

igvc/src/global_mapper/main.cpp
```diff
@@ -11,56 +11,89 @@
...
+double probability_thresh;
 
-void icp_transform(pcl::PointCloud<pcl::PointXYZ>::Ptr input)
+std::list<pcl::PointCloud<pcl::PointXYZRGB>::Ptr> pointcloud_list;
+int queue_window_size;
+
+void icp_transform(pcl::PointCloud<pcl::PointXYZRGB>::Ptr input)
 {
-  if (!firstFrame)
+  if (pointcloud_list.size() < queue_window_size)
```
JasonGibson274: comparison between signed and unsigned

igvc/src/global_mapper/main.cpp
```diff
@@ -11,56 +11,89 @@
...
     }
-    new_points.header.frame_id = "/odom";
-    _pointcloud_incremental_pub.publish(new_points);
+
+    (*local_pc).header.frame_id = "/odom";
+    _pointcloud_prob_pub.publish(*local_pc);
+
+    // Remove first pointcloud
+    pointcloud_list.pop_front();
     std::cout << "Map size: " << global_map->size() << std::endl;
```
JasonGibson274: Can we remove this?

JasonGibson274:
I think this is a good step in the right direction but I would like to go to actual global mapping after this. Also @dpattison3 can you do a review?

jzheng84:
We have looked into this, there are a ton of libraries for mapping but they are all for specific uses (indoor only, lidar only, camera only, etc). We should definitelu look to use the algorithms being implemented in these libraries, although it will messy to put things together.

JasonGibson274:
Long term project for sure

PR #257, opened 2017-11-16
rviz plugins setup

### LINKED ISSUE #250: Set up catkin for rviz plugins
Setup and example for freshmen to start working on their plugins as freshmen projects.

barulicm:
Adds the igvc_rviz_plugins package with catkin setup for rviz plugins.
Adds ExamplePanel which displays the speeds published on the /encoders topic.
Fixes #250

### DIFF
--- a/igvc_rviz_plugins/CMakeLists.txt
+++ b/igvc_rviz_plugins/CMakeLists.txt
@@ -1,0 +1,54 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(igvc_rviz_plugins)
+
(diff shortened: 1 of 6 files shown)


barulicm:
And @JasonGibson274 , by the way, the last attempt at rviz plugins in this repo still exists at igvc/src/rviz_plugins. I haven't looked to see if any of those panels have useful implementations in them, so you may want to copy the QT parts of that code into this new package.

JasonGibson274:
Will do

PR #261, opened 2017-11-19
Somewhat nice randomly generated test textures

atuttle7:
by Andrew Tuttle and Keshav Kolur


JasonGibson274:
Does this run in Linux? I am getting tabs and spaces in the same file.

JasonGibson274:
This pull request looks ready to merge except that a gazebo material needs to be created in order to actually use what was built. It will take about 20 minutes to set that up so next meeting just grab me and we will do that. Also I want to check the lidar bounding box while running just for sanity sake but code wise everything seems fine.

Issue #255, opened 2017-11-13
Rviz plugin for sensor status


JasonGibson274:
Panel that shows each sensor and if it is publishing data to its topics.

KishoreMayank:
Aaron McDaniel
Soohyun Kim
Mayank Kishore

soohyun-kim:
Gazebo and RVIZ response to Sensor Panels seems to be working; only need the timer component and I think we'll be finished

JasonGibson274:
Resolved

PR #267, opened 2017-11-29
fixed issue #255

AaronMcDaniel:
We implemented the sensor panel. It works as far as we can tell.


igvc_rviz_plugins/src/sensor_panel.cpp
```diff
@@ -0,0 +1,171 @@
+#include <igvc_rviz_plugins/sensor_panel.h>
+#include <pluginlib/class_list_macros.h>
+#include <QVBoxLayout>
+//#include <std>
+//#include "rjTimer.h"
```
JasonGibson274: remove unused headers

igvc_rviz_plugins/src/sensor_panel.cpp
```diff
@@ -0,0 +1,171 @@
...
+namespace igvc_rviz_plugins
+{
+sensor_panel::sensor_panel(QWidget *parent) : rviz::Panel(parent)  // Base class constructor
+{
+  // Panels are allowed to interact with NodeHandles directly just like ROS nodes.
+  ros::NodeHandle nh;
+
+  // Initialize a label for displaying some data
+
+  labels[0] = new QLabel("IMU: Placehold\n");
```
JasonGibson274: see previous comment about converting to map

igvc_rviz_plugins/include/igvc_rviz_plugins/sensor_panel.h
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Declare any ROS callbacks you need here. Be sure to create a parameter for any UI elements you need to update.
+   * @param msg The ROS message that triggers this callback.
+   * @param label A QT label whose text we will update based on the message contents.
+   */
+  void imu_callback(const sensor_msgs::ImuConstPtr &msg, QLabel *label);
+  void lidar_callback(const sensor_msgs::PointCloud2ConstPtr &msg, QLabel *label);
+  void gps_callback(const sensor_msgs::ImuConstPtr &msg, QLabel *label);
+  void camCenter_callback(const sensor_msgs::ImageConstPtr &msg, QLabel *label);
```
JasonGibson274: cam_center_callback?

igvc_rviz_plugins/src/sensor_panel.cpp
```diff
@@ -0,0 +1,171 @@
...
+  labels[2] = new QLabel("CamCenter: Placehold\n");
+  labels[3] = new QLabel("CamLeft: Placehold\n");
+  labels[4] = new QLabel("CamRight: Placehold\n");
+  labels[5] = new QLabel("GPS: Placehold");
+
+  /*
+  *Makes timer that repeats every interval seconds
+  *When timer goes off, it calls the timer method
+  */
+  // const int interval = 500;
```
JasonGibson274: remove unused line?

igvc_rviz_plugins/src/sensor_panel.cpp
```diff
@@ -0,0 +1,171 @@
...
+  {
+    layout->addWidget(labels[i]);
+  }
+  setLayout(layout);
+  timer();  // inits to disabled
+}
+
+void sensor_panel::imu_callback(const sensor_msgs::ImuConstPtr &msg, QLabel *label)
+{
+  label->setText("IMU: Enabled" /*.c_str()*/);
```
JasonGibson274: remove commented out code

igvc_rviz_plugins/src/sensor_panel.cpp
```diff
@@ -0,0 +1,171 @@
...
+  label->setText("GPS: Enabled");
+  isActive[5] = true;
+}
+
+/*
+*Sets all of the activity statuses to disabled, and sets labels to diabled if already disabled
+*/
+void sensor_panel::reset_labels()
+{
+  for (int i = 0; i < NUMSENSORS; i++)
```
JasonGibson274: for each loop?

igvc_rviz_plugins/src/sensor_panel.cpp
```diff
@@ -0,0 +1,171 @@
...
+      set_label(i, isActive[i]);  // set to false on second time
+    }
+    isActive[i] = false;  // clear all statuses
+  }
+}
+
+/*
+*Sets the label at the i position to active status of b
+*/
+void sensor_panel::set_label(int i, bool b)
```
JasonGibson274: A map will make this better

igvc_rviz_plugins/src/sensor_panel.cpp
AaronMcDaniel: Done. Uncommitted.

igvc_rviz_plugins/include/igvc_rviz_plugins/sensor_panel.h
soohyun-kim: camel typing ftw (fixed)

AaronMcDaniel: Done. Pushed

igvc_rviz_plugins/src/sensor_panel.cpp
AaronMcDaniel: Pushed fix.

igvc_rviz_plugins/src/sensor_panel.cpp
AaronMcDaniel: Done. Pushed.

igvc_rviz_plugins/src/sensor_panel.cpp
AaronMcDaniel: Done. Pushed.

igvc_rviz_plugins/src/sensor_panel.cpp
AaronMcDaniel: Done. Pushed.

igvc_rviz_plugins/src/sensor_panel.cpp
AaronMcDaniel: Done. Pushed

igvc_rviz_plugins/include/igvc_rviz_plugins/sensor_panel.h
```diff
@@ -0,0 +1,77 @@
...
+  void camLeft_callback(const sensor_msgs::ImageConstPtr &msg, QLabel *label);
```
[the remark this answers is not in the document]
soohyun-kim: Done.

igvc_rviz_plugins/src/sensor_panel.cpp
AaronMcDaniel: fixed. Pushed. The method in the header needs to take in a std::string now instead of an int

igvc_rviz_plugins/src/sensor_panel.cpp
AaronMcDaniel: If I use a fro each loop and i have to access the labels stored in the map in the set_label() method. How can I figure out which QLabel each bool in isActive corresponds to?

AaronMcDaniel: I think that these two data structures would be easiest to deal with if they had the same indexing structure. So I think unless we store the bools in a map with corresponding keys to the labels then an array would be better to use.

igvc_rviz_plugins/include/igvc_rviz_plugins/sensor_panel.h
```diff
@@ -0,0 +1,77 @@
...
+#include <sensor_msgs/Image.h>
+#include <sensor_msgs/Imu.h>
+#include <sensor_msgs/NavSatFix.h>
+#include <sensor_msgs/PointCloud2.h>
+#include <QLabel>
+#include <QObject>
+#include <QTimer>
+#include <string>
+
+#define INTERVAL 250
```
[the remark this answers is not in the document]
AaronMcDaniel: fixed. Pushed.

igvc_rviz_plugins/include/igvc_rviz_plugins/sensor_panel.h
```diff
@@ -0,0 +1,77 @@
...
+   */
+  ros::Subscriber imu_sub;
+  ros::Subscriber lidar_sub;
+  ros::Subscriber gps_sub;
+  ros::Subscriber camCenter_sub;
+  ros::Subscriber camLeft_sub;
+  ros::Subscriber camRight_sub;
+
+  QTimer *sensor_timer;
+  QLabel *labels[NUMSENSORS];  // list of pointers to labels
```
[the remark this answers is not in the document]
AaronMcDaniel: fixed. causes an issue with some methods. Still needs to be fixed

igvc_rviz_plugins/include/igvc_rviz_plugins/sensor_panel.h
```diff
@@ -0,0 +1,77 @@
...
+  bool isActive[NUMSENSORS];   // list of activity for sensors
+
+  /**
+   * Declare any ROS callbacks you need here. Be sure to create a parameter for any UI elements you need to update.
+   * @param msg The ROS message that triggers this callback.
+   * @param label A QT label whose text we will update based on the message contents.
+   */
+  void imu_callback(const sensor_msgs::ImuConstPtr &msg, QLabel *label);
+  void lidar_callback(const sensor_msgs::PointCloud2ConstPtr &msg, QLabel *label);
+  void gps_callback(const sensor_msgs::ImuConstPtr &msg, QLabel *label);
```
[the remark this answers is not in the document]
AaronMcDaniel: Fixed this. There were also no callbacks for left or right camera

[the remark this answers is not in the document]
soohyun-kim: Please note that gps_callback cannot be used, as that function already exists within an in-use namespace.  Use gps2_callback

igvc_rviz_plugins/src/sensor_panel.cpp
soohyun-kim: I concur; I fail to see how a for-each loop would be any more convenient, considering that additional code elements are necessary to iterate over a map

JasonGibson274: This comment was centered around getting rid of the sensor number being hardcoded.

igvc/src/rviz_plugins/speed_panel.cpp
```diff
@@ -25,7 +25,7 @@ SpeedPanel::SpeedPanel(QWidget* parent) : rviz::Panel(parent)
 {
   speedometer = new Speedometer(parent);
   QHBoxLayout* speedDisp = new QHBoxLayout;
-  speedDisp->addWidget(new QLabel("SPEEDOMETER"));
+  speedDisp->addWidget(new QLabel("THIS IS A TESET"));
```
JasonGibson274: Can this be reverted?

igvc_rviz_plugins/include/igvc_rviz_plugins/sensor_panel.h
```diff
@@ -0,0 +1,71 @@
...
+   * the lifetime of the widget.
+   */
+  ros::Subscriber imu_sub;
+  ros::Subscriber lidar_sub;
+  ros::Subscriber gps_sub;
+  ros::Subscriber cam_center_sub;
+
+  QTimer *sensor_timer;
+  std::map<std::string, QLabel *> labels;// map of pointers to labels
+  bool *isActive; // list of activity for sensors
```
JasonGibson274: Either convert this to a strut so we can see if a label is active or convert the map to having a pair with the QLabel and the bool. I do not like maintaining separate data structures with related data.

igvc_rviz_plugins/include/igvc_rviz_plugins/sensor_panel.h
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
```
JasonGibson274: I think we can remove these comments or make them useful, personal discretion here.

igvc_rviz_plugins/src/sensor_panel.cpp
```diff
@@ -0,0 +1,109 @@
...
+ */
+namespace igvc_rviz_plugins
+{
+sensor_panel::sensor_panel(QWidget *parent) : rviz::Panel(parent)  // Base class constructor
+{
+  // Panels are allowed to interact with NodeHandles directly just like ROS nodes.
+  ros::NodeHandle nh;
+
+  // Initialize all labels for displaying some data
+  labels["IMU"] = new QLabel("IMU: Placehold\n");
```
JasonGibson274: Great place for a shared pointer

igvc_rviz_plugins/src/sensor_panel.cpp
```diff
@@ -0,0 +1,109 @@
...
+  labels["Lidar"] = new QLabel("Lidar: Placehold\n");
+  labels["Center Camera"] = new QLabel("Center Camera: Placehold\n");
+  labels["GPS"] = new QLabel("GPS: Placehold\n");
+
+  isActive = new bool [labels.size()];
```
JasonGibson274: shared pointer

igvc_rviz_plugins/src/sensor_panel.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+  /*
+  *Makes timer that repeats every interval seconds
+  *When timer goes off, it calls the timer method
+  */
+  sensor_timer = new QTimer();
```
JasonGibson274: shared pointer

igvc_rviz_plugins/include/igvc_rviz_plugins/sensor_panel.h
JasonGibson274: Applies to all comments

igvc_rviz_plugins/src/sensor_panel.cpp
AaronMcDaniel: I tried using shared pointers and ended up getting some boost errors. This may take a bit, but once I fix this everything should be working fine.

JasonGibson274: If shared pointers are giving issues then just leave them as is.

AaronMcDaniel: Ok, I'll get rid of them. I also realized that QT doesn't like it when I use shared pointers either.

igvc_rviz_plugins/include/igvc_rviz_plugins/sensor_panel.h
AaronMcDaniel: I opted for a struct, because the string in the map made resetting the labels much easier.

igvc_rviz_plugins/include/igvc_rviz_plugins/sensor_panel.h
AaronMcDaniel: Im fixing the comments still

JasonGibson274:
I get a lot of flipping back and forth between enabled and disabled when publishing the topics. Can you prevent that?

AaronMcDaniel:
Ya, I will work on that. I got the same issue but when we tried it on Soohyun's laptop it worked smoothly. I will see what I can do. We may just need to slow the timer speed down.

AaronMcDaniel:
I fixed the label issue. I was restarting the timer manually even though it was set to restart automatically.

JasonGibson274:
Is it still timing out for you?

JasonGibson274:
If I publish to the topic once it never goes back to disabled

AaronMcDaniel:
My bad. The extra call to the timer was also the way the timer started initially. I need to keep making changes.

AaronMcDaniel:
For some reason the timer is sending the timeout signal 2 times really quickly every time it times out. And as far as the documentation goes there should only be one signal sent each time.

AaronMcDaniel:
I think the bools were being reinitialized each frame. Perhaps the constructor gets called frequently? Idk exactly what the problem was but I found a work around.

JasonGibson274:
Functionality wise this is done, but is it possible to make it more graphic i.e. large boxes with the sensor names that are green if enabled and red disabled

AaronMcDaniel:
Sure. I just found the documentation for it. I'll put the text for each sensor in its own colored box, and probably bold the text so it is more legible

AaronMcDaniel:
I'm not the best with aesthetics, but that was super easy so let me know if you want it to look any different

JasonGibson274:
AMAZING. LGTM!

PR #278, opened 2018-01-21
Gazebo ground texture mapping

atuttle7:
Sorted out some github branching issues. This should be the same code as was in the old pull request.


atuttle7:
Why is github hard; i'll fix this one too

PR #281, opened 2018-01-24
Pathplanner

JasonGibson274:
Still a work in progress.
Update the path planner to have parameters for all important variables. Fixed issue where the 3d mapping resulted in inaccurate distances to obstacles.


igvc/src/pathplanner/main.cpp
```diff
@@ -106,25 +106,37 @@ int main(int argc, char** argv)
...
   search_problem.Octree->setInputCloud(search_problem.Map);
-  search_problem.GoalThreshold = 1.0;
-  search_problem.Threshold = 0.5;
-  search_problem.Speed = 1.0;
-  search_problem.Baseline = baseline;
+
+  if(!pNh.hasParam("goal_threshold") || !pNh.hasParam("threshold") || !pNh.hasParam("speed") || !pNh.hasParam("baseline") || !pNh.hasParam("minimum_omega") || !pNh.hasParam("maximum_omega") || !pNh.hasParam("delta_omega") || !pNh.hasParam("point_turns_enabled") || !pNh.hasParam("reverse_enabled") || !pNh.hasParam("max_obstacle_delta_t") || !pNh.hasParam("alpha") || !pNh.hasParam("beta") || !pNh.hasParam("bounding_distance")) {
+    ROS_ERROR_STREAM("path planner does not have all required parameters");
+    ROS_ERROR_STREAM("path planner does not have all required parameters");
+    ROS_ERROR_STREAM("path planner does not have all required parameters");
```
barulicm: Duplicate error messages?

JasonGibson274: I should do a ros rate that continues to publish or set default values.

JasonGibson274:
Not leaving a comment is a bad habit of mine. I will start doing that.

igvc/src/pathplanner/main.cpp
barulicm: It's also OK to just have the node shut itself down if it's started without all of the parameters it needs.

JasonGibson274: That’s actually a much better idea. I will do that.

igvc/src/pathplanner/main.cpp
```diff
@@ -115,8 +115,7 @@ int main(int argc, char** argv)
 
   if(!pNh.hasParam("goal_threshold") || !pNh.hasParam("threshold") || !pNh.hasParam("speed") || !pNh.hasParam("baseline") || !pNh.hasParam("minimum_omega") || !pNh.hasParam("maximum_omega") || !pNh.hasParam("delta_omega") || !pNh.hasParam("point_turns_enabled") || !pNh.hasParam("reverse_enabled") || !pNh.hasParam("max_obstacle_delta_t") || !pNh.hasParam("alpha") || !pNh.hasParam("beta") || !pNh.hasParam("bounding_distance")) {
     ROS_ERROR_STREAM("path planner does not have all required parameters");
-    ROS_ERROR_STREAM("path planner does not have all required parameters");
-    ROS_ERROR_STREAM("path planner does not have all required parameters");
+    ros::shutdown();
```
barulicm: You should just return from main() here, rather than calling shutdown(). ros::shutdown(), in fact, does not actually terminate the program. It's responsible for cleaning up a lot of the behind-the-scenes resources in ROS, but this functionality is taken care of when the NodeHandle is destructed.
From the ROS documentation:
"It is generally not necessary to call this function, as the node will automatically shutdown when all NodeHandles destruct. However, if you want to break out of a spin() loop explicitly, this function allows that." [[ref](http://docs.ros.org/api/roscpp/html/namespaceros.html#a2d29113188d411e78f2c221aa7a7473e)]

Issue #252, opened 2017-11-13
Rviz plugin for waypoint


JasonGibson274:
I would like to be able to see relative heading to the waypoint and how far away it is. Follow the directions in igvc/src/igvc_rviz_plugins for information

Remulus2006:
Harsh Jain

JasonGibson274:
* [ ] distance to waypoint
* [ ] angle to waypoint (i.e. and arrow pointing to where it is)
* [ ] status of the waypoint

bvaughn8:
I'll make a pull request now, the panel displays the distance to the waypoint and the angle (as a number, not graphically).

PR #282, opened 2018-01-25
Waypoint panel #252

bvaughn8:
New pull request that doesn't include a bunch of messed up files.


bvaughn8:
I'm currently getting an error when I run rviz that it cannot find rviz::Panel...trying to fix it

igvc_rviz_plugins/plugin_description.xml
```diff
@@ -15,4 +15,12 @@
         </description>
     </class>
 
+    <class name="igvc_rviz_plugins/WaypointPanel"
+           type="igvc_rviz_plugins::WaypointPanel"
+           base_class_type="rviz::Panel"
```
barulicm: You are missing a ">" to finish the opening tag of the _class_ element here. This is causing the XML file to be malformed, causing rviz to be unable to read it.

bvaughn8:
@JasonGibson274 where do I add to cmake file?

JasonGibson274:
Line like https://github.com/RoboJackets/igvc-software/blob/master/igvc_rviz_plugins/CMakeLists.txt#L34 **and**
https://github.com/RoboJackets/igvc-software/blob/master/igvc_rviz_plugins/CMakeLists.txt#L38

igvc_rviz_plugins/src/WaypointPanel.cpp
```diff
@@ -0,0 +1,93 @@
...
+    robot_y = msg->pose.pose.position.y;
+
+    double qx = msg->pose.pose.orientation.x;
+    double qy = msg->pose.pose.orientation.y;
+    double qz = msg->pose.pose.orientation.z;
+    double qw = msg->pose.pose.orientation.w;
+    double siny = 2.0 * (qw * qz + qx * qy);
+    double cosy = 1.0 - 2.0 * (qy * qy + qz * qz);
+    // double sinr = 2.0 * (qw * qx + qy * qz);
+    // double cosr = 1.0 - 2.0 * (qx * qx + qy * qy);
```
JasonGibson274: delete commented out lines if unused

JasonGibson274:
Looks good to me

igvc_rviz_plugins/src/WaypointPanel.cpp
```diff
@@ -0,0 +1,56 @@
...
+    setLayout(layout);
+}
+
+void WaypointPanel::waypoint_callback(const geometry_msgs::PointStampedConstPtr &msg, QLabel *label) {
+    way_x = msg->point.x - robot_x;
+    way_y = msg->point.y - robot_y;
+}
+
+void WaypointPanel::robot_position_callback(const nav_msgs::OdometryConstPtr &msg, QLabel *label) {
+    double distance = sqrt(pow(way_x, 2) + pow(way_y, 2));
```
barulicm: Consider replacing with a call to [std::hypot](http://en.cppreference.com/w/cpp/numeric/math/hypot)

igvc_rviz_plugins/src/WaypointPanel.cpp
```diff
@@ -0,0 +1,56 @@
...
+    robot_y = msg->pose.pose.position.y;
+
+    double qx = msg->pose.pose.orientation.x;
+    double qy = msg->pose.pose.orientation.y;
+    double qz = msg->pose.pose.orientation.z;
+    double qw = msg->pose.pose.orientation.w;
+    double siny = 2.0 * (qw * qz + qx * qy);
+    double cosy = 1.0 - 2.0 * (qy * qy + qz * qz);
+
+    double yaw = atan2(siny, cosy) * 180 / M_PI;
```
barulicm: Consider using a library for this. For example:
```
double roll, pitch, yaw;
tf::Matrix3x3(q).getRPY(roll, pitch, yaw);
```
This will improve readability as it makes it clear what you're trying to do, and makes it less likely that someone will accidentally mess this math up in the future.

JasonGibson274:
I would say make the changes. Getting used to tf matrixes is always good.

igvc_rviz_plugins/src/WaypointPanel.cpp
```diff
@@ -0,0 +1,61 @@
...
+    double qz = msg->pose.pose.orientation.z;
+    double qw = msg->pose.pose.orientation.w;
+
+    tf::Quaternion q(qx, qy, qz, qw);
+    tf::Matrix3x3 m(q);
+    double roll, pitch, yaw;
+    m.getRPY(roll, pitch, yaw);
+
+
+    //double yaw = atan2(siny, cosy) * 180 / M_PI;
```
barulicm: Try to avoid adding "dead code." (Commented out code). It's a good habit to build early. I've seen instances where huge long blocks of code are commented out so that another huge long alternative block of code below can be used instead.

igvc_rviz_plugins/src/WaypointPanel.cpp
```diff
@@ -0,0 +1,61 @@
...
+}
+
+void WaypointPanel::waypoint_callback(const geometry_msgs::PointStampedConstPtr &msg, QLabel *label) {
+    way_x = msg->point.x - robot_x;
+    way_y = msg->point.y - robot_y;
+}
+
+void WaypointPanel::robot_position_callback(const nav_msgs::OdometryConstPtr &msg, QLabel *label) {
+    double distance = std::hypot(way_x, way_y);
+    double angle = atan2(way_y, way_x) * 180 / M_PI;
```
barulicm: If you can, give this a more descriptive name. Here, that could be as simple as `globalAngle` or `angleBetweenPoints`. Pick a name you like that would explain to a new reader of this code that this is not just any angle but an angle measurement that means something specific.

barulicm:
Code looks good. CI failed, but it doesn't look code-related. I'm triggering a rebuild to double check.

PR #295, opened 2018-02-28
Vision

JasonGibson274:
This PR should not be merged until 
* [x] pointcloud format is pointcloud2
* [x] CI no longer fails

This contains all of the changes necessary for running the NN vision. It is written in python but I would like to have this merged in soon so that we no longer need to merge this branch in every single time we go testing. 

Also removes all of the old vision files.


JasonGibson274:
After some investigation we are not able to install using pip since it is not installed

JasonGibson274:
@jgkamat  I think I resolved the issue by adding pip to be installed in the docker file, but it seems to be pulling the cached image and I still get the same error. Is there a way to trigger a clean build of the image or am I still missing something?

jgkamat:
An easy way would be to add a blank line to [any of these files](https://github.com/RoboJackets/igvc-software/blob/master/config.docif#L81-L86), and that should trigger a clean rebuild. The dockerfile isn't actually in the SHA files, so a better way would be adding the dockerfile to that list.

JasonGibson274:
@dpattison3 just a final review and we can officially merge the only part of IGVC we have solved.

igvc/src/cnn/cnn.py
```diff
@@ -0,0 +1,173 @@
...
+        image_pl = self.graph.get_tensor_by_name('Placeholder_1:0')
+        softmax = self.graph.get_tensor_by_name('Validation/decoder/Softmax:0')
+
+        output = self.sess.run([softmax], feed_dict={image_pl: im})
+        shape = im.shape
+        output_image = output[0][:, 1].reshape(shape[0], shape[1])
+
+        threshold = 0.5
+        im_threshold = output_image > threshold
+        world_points = self.world_point_array[im_threshold].tolist()
```
dpattison3: We convert this numpy array to a regular list, then convert it back on line 151.

igvc/src/cnn/cnn.py
```diff
@@ -0,0 +1,173 @@
...
+        cloud_msg.height = 1
+        cloud_msg.width = len(world_points)
+        cloud_msg.fields = [
+            PointField('x', 0, PointField.FLOAT32, 1),
+            PointField('y', 4, PointField.FLOAT32, 1),
+            PointField('z', 8, PointField.FLOAT32, 1)]
+        cloud_msg.is_bigendian = False
+        cloud_msg.point_step = 12
+        cloud_msg.row_step = 3 * len(world_points)
+        cloud_msg.data = np.asarray(world_points, np.float32).tostring()
```
dpattison3: I'm not exactly sure, but is it possible to store the string array instead of list of points (when initializing earlier)? If we can, that might be slightly faster (possibly negligibly so but not sure).

barulicm:
As a side note, I would recommend separating really big PRs like this for the future. This both:
1. Adds the new CNN code
2. Removes a lot of old CV code

In fact, most of the files touched by the PR are for the second task, but this task is only documented in one of the commit messages.
When possible, add new code, continue testing, then remove old code.

All in all though, impressive work. I'll leave the actual code reviews to the Python experts.

PR #293, opened 2018-02-25
Create node to compare requested vs. actual motor speeds

evan10s:
With the zillions of commits, I think `cherry-pick`ing the relevant commits or doing a squash merge are the best options here

PR #296, opened 2018-03-01
Node to compare real vs target speed 

AaronMcDaniel:
Files copied from old pull request #293


igvc/src/speed_compare/main.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <chrono>
+#include <fstream>
+
+bool enabled = false;
+std::ofstream file;
+float target_left_vel;
+float target_right_vel;
+ros::Time start_time;
+
+void enabled_callback(const std_msgs::BoolConstPtr& msg)
```
JasonGibson274: Just a note here. On the actual robot /robot_enabled will be continually published (10Hz I think) so this will stream output. I do not think that is a problem just a note here.

igvc/src/speed_compare/main.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+void motors_callback(const igvc_msgs::velocity_pair& msg)
+{
+  if (enabled)
+  {
+    target_left_vel = msg.left_velocity;
+    target_right_vel = msg.right_velocity;
+  }
+  else
```
JasonGibson274: I would remove the logging here since it is just a repeat of the above callback information as well

igvc/src/speed_compare/main.cpp
```diff
@@ -0,0 +1,77 @@
...
+  {
+    ROS_INFO_STREAM("Data stored");
+    ros::Duration time_diff = msg.header.stamp - start_time;
+    file << time_diff.sec << "." << time_diff.nsec << ",";
+    file << msg.left_velocity << "," << msg.right_velocity << ",";
+    file << target_left_vel << "," << target_right_vel << std::endl;
+  }
+  else
+  {
+    ROS_INFO_STREAM("Robot disabled, not writing data");
```
JasonGibson274: I would remove this logging since it is just a repeat of the robot enabled callback message

igvc/src/speed_compare/main.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+  start_time = ros::Time::now();
+
+  ros::Subscriber enabled_sub = nh.subscribe("/robot_enabled", 1, enabled_callback);
+  ros::Subscriber motors_sub = nh.subscribe("/motors", 1, motors_callback);
+  ros::Subscriber encoders_sub = nh.subscribe("/encoders", 1, encoders_callback);
+
+  std::string location;
+  pNh.getParam("file", location);
+  ROS_INFO_STREAM(location);
```
JasonGibson274: can you add context on what this location is to the message. For example saving data to file at [location[

igvc/launch/speed_compare.launch
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0"?>
+
+<!-- map.launch -->
+<launch>
+
+    <!-- Planner -->
+    <!--<include file="$(find igvc)/sandbox/speed_data.csv" />-->
```
JasonGibson274: remove if not used

.gitignore
```diff
@@ -30,6 +30,9 @@
 #rosdoc_lite
 /doc/
 
+#Speed_compare
```
JasonGibson274: Might want to create a folder instead of just ignoring a single file. Then I can create multiple and store them there.

igvc/src/speed_compare/main.cpp
AaronMcDaniel: This was just for testing, good catch

sandbox/speed_compare/speed_data.csv
```diff
@@ -0,0 +1,51 @@
+timestamp, left_velocity, right_velocity, target_left_velocity, target_right_velocity
```
JasonGibson274: delete this file

AaronMcDaniel:
Good catch, I didn't notice that at all. Its gone now though

PR #300, opened 2018-03-03
Waypoint data recorder

bvaughn8:
Transcribes data from /waypoint to a csv specified in the launch file


igvc/src/waypoint_data_recorder/main.cpp
```diff
@@ -0,0 +1,31 @@
+#include <geometry_msgs/PointStamped.h>
+#include <ros/ros.h>
+#include <iostream>
+#include <fstream>
+
+std::string fileName;
+
+void waypoint_callback(const geometry_msgs::PointStampedConstPtr& msg)
+{
+	std::ofstream datacsv;
```
barulicm: formatting issue, this line's indent doesn't match the rest of this file.

igvc/src/waypoint_data_recorder/main.cpp
```diff
@@ -0,0 +1,31 @@
...
+    datacsv.open (fileName, std::ios_base::app | std::ios_base::out);
```
barulicm: Is there a reason you've chosen to re-open the file everytime the waypoint callback is hit?

bvaughn8: I did it this way to make sure the file closes properly since I have to stop the node using ctrl+c. If I can close the file in the main method and it still works properly, I can make that change. I just didn't think it would close correctly. To address the question about the point, I used this to transcribe the data being published in /waypoint which is different from the data in the waypoint csv. With this data, I am working with @cchoate6 to create a filter that makes the distance to the waypoint less wobbly/more consistent.

barulicm: Thanks! Totally forgot that waypoint_source applies a frame transformation. 

As for your clean shutdown concern. ROS handles this for you. Ctrl-C simply sends an interrupt signal (SIGINT) to your program. ROS handles this signal for you to cleanly shutdown its networking resources. ros::spin() will return when ros has handled that signal. Any objects in the main method will safely destruct when main then returns. So, in this case, an ofstream object in the main method will close itself cleanly in its destructor.

JasonGibson274:
We are probably going to move this to the bag file utility package that I am writing right now once that is done.

igvc/src/waypoint_source/main.cpp
```diff
@@ -16,6 +16,9 @@ geometry_msgs::PointStamped current_waypoint;
 geometry_msgs::Point map_origin;
 std::mutex current_mutex;
 
+float waypointFilterXArray[5] = {3.4e38, 3.4e38, 3.4e38, 3.4e38, 3.4e38};
```
samr28: Maybe use `std::numeric_limits<float>::max();` instead of `3.4e38`

igvc/src/waypoint_source/main.cpp
```diff
@@ -131,6 +138,26 @@ void originCallback(const sensor_msgs::NavSatFixConstPtr& msg)
   }
 }
 
+
+float waypointFilter(float newEntry, float filteringArray[]) {
+  float sum = 0.0;
+  if (filteringArray[0] == 3.8e38) {
```
samr28: Can it ever hit this case? Max size for a float is 3.4e38

Issue #298, opened 2018-03-02
add elevation to gazebo


JasonGibson274:
see the instructions here work before going to the gazebo OEM files.

https://github.com/AS4SR/general_info/wiki/Creating-Heightmaps-for-Gazebo

AaronMcDaniel:
@JasonGibson274 Can you link me that .materials file?

AaronMcDaniel:
Found it in #299

Alescontrela:
Currently being implemented, therefore not a bug

PR #303, opened 2018-03-29
Elevated Terrain in Gazebo

AaronMcDaniel:
Made terrain object for gazebo which has elevation as a fix for issue #298. Includes minor changes to URDF to get the robot to move.


gazebo/igvc_control/launch/gazebo.launch
```diff
@@ -3,9 +3,12 @@
 <launch>
   <env name="GAZEBO_MODEL_PATH" value="$(find igvc_description)"/>
   <arg name="namespace" default="igvc"/>
+  <!-- Switch with commented line below for better performance -->
```
JasonGibson274: I think here you mean uncomment below for terrain within the world

gazebo/igvc_description/urdf/models/grass_plane/model.config
```diff
@@ -4,7 +4,6 @@
   <name>Grass Plane</name>
   <version>1.0</version>
   <sdf version="1.2">model-1_2.sdf</sdf>
-  <sdf version="1.3">model-1_3.sdf</sdf>
```
JasonGibson274: I do not think this is a breaking change since we should all be above this version of sdf but I was just wondering if there was a reason this was deleted.

AaronMcDaniel: We have no model-1_3.sdf file. I think we talked about this before, but it wasn't doing anything anyway

gazebo/igvc_control/launch/gazebo.launch
AaronMcDaniel: Ya, I'll phase that better

gazebo/igvc_description/urdf/models/grass_plane/model.config
JasonGibson274: yep yep, that is my bad.

PR #313, opened 2018-04-18
Firmware

JasonGibson274:
This PR encompasses the necessary firmware and other hardware related changes to switching to Jessi. A new motor controller protocol was implemented. The URDF file was updated to represent Jessi. The IMU was changed from the Razor AHS to the [FSM-9](https://www.hillcrestlabs.com/products/fsm-9). Before this is merged three things must happen

* [ ] A first year member must fully understand and approve this PR
* [ ] An experienced member must review the PR
* [x] The IMU has to be properly calibrated. 

First year members should start reviewing the code now but ignore the imu file.


gazebo/igvc_description/urdf/jessi.urdf
```diff
@@ -275,6 +275,8 @@
       <mechanicalReduction>1</mechanicalReduction>
     </actuator>
   </transmission>
+
+
```
evan10s: Remove the extra line of whitespace here

gazebo/igvc_description/urdf/jessi.urdf
```diff
@@ -331,7 +333,7 @@
   <joint name="imu_to_front_ball" type="floating">
     <parent link="imu"/>
     <child link="front_ball"/>
-    <origin xyz=".39 0 -.53" rpy="0 0 0" />
+    <origin xyz=".39 -0 -.53" rpy="0 0 0" />
```
evan10s: There shouldn't be a negative sign in front of the middle entry (`-0`) in the value of the `xyz` attribute

igvc/launch/imu.launch
```diff
@@ -1,4 +1,4 @@
 <!-- imu.launch -->
 <launch>
-    <node name="imu" pkg="igvc" type="imu"/>
+    <node name="imu" pkg="igvc" type="imu" output="screen" />
```
evan10s: If `output="screen"` was added for debugging, it should be removed before merging

igvc/launch/localization.launch
```diff
@@ -70,12 +70,13 @@
       <param name="imu0" value="/imu"/>
       <!-- fused parameters same as above -->
       <rosparam param="imu0_config">[false, false, false,
-                                 true,  true,  true,
```
evan10s: Is there documentation anywhere for what this array of boolean values means?  Perhaps that should be included in a comment next to this value

igvc/launch/sensors.launch
```diff
@@ -11,7 +11,15 @@
     <node name="light_controller" pkg="igvc" type="light_controller" output="screen"/>
 
     <!-- MOTOR CONTROLLER -->
-    <node name="motor_controller" pkg="igvc" type="motor_controller" />
+    <node name="motor_controller" pkg="igvc" type="motor_controller" output="screen">
+          <param name="p_l" type="double" value="12.0"/>
+          <param name="p_r" type="double" value="12.0"/>
+          <param name="d_l" type="double" value="0.5"/>
+          <param name="d_r" type="double" value="0.5"/>
+          <param name="i_r" type="double" value="0.0"/>
```
evan10s: When we were testing the motors last Friday afternoon, I noticed that the motor speed overshot the target slightly for full forward power and didn't quite reach the target for full backward power.  Setting the integral constant to something small (I would guess on the order of .001) would help fix these small errors.

igvc/src/imu/main.cpp
```diff
@@ -91,29 +99,56 @@ int main(int argc, char** argv)
...
     {
-      msg.linear_acceleration.x = stof(accelTokens[0]);
-      msg.linear_acceleration.y = stof(accelTokens[1]);
-      msg.linear_acceleration.z = stof(accelTokens[2]);
+      double cur_x = stof(accelTokens[0]);
+      double cur_y = stof(accelTokens[1]);
+      double cur_z = stof(accelTokens[2]);
       msg.linear_acceleration_covariance = { 0.04, 1e-6, 1e-6, 1e-6, 0.04, 1e-6, 1e-6, 1e-6, 0.04 };
 
+      int num = 500;
```
evan10s: It looks like this variable is setting a threshold to form some kind of buffer / moving average for the x/y/z accel readings.  If that's the case, it should have a name that's more descriptive than `num`.

igvc/src/imu/main.cpp
```diff
@@ -91,29 +99,56 @@ int main(int argc, char** argv)
...
+      msg.linear_acceleration.z = cur_z - z_avg;
+
+
       msg.angular_velocity.x = stof(gyroTokens[0]);
       msg.angular_velocity.y = stof(gyroTokens[1]);
       msg.angular_velocity.z = stof(gyroTokens[2]);
       msg.angular_velocity_covariance = { 0.02, 1e-6, 1e-6, 1e-6, 0.02, 1e-6, 1e-6, 1e-6, 0.02 };
 
-      float yaw = HALF_TO_FULL_CIRCLE_ANGLE(-(stof(rpyTokens[0]) + 90)) * DEG_TO_RAD;
+      float yaw = HALF_TO_FULL_CIRCLE_ANGLE(-(stof(rpyTokens[0]) - 100)) * DEG_TO_RAD;
```
evan10s: Why was `+ 90` changed to `- 100`?  If there's some sort of tuning going on here, I think it should be written similar to `- 90 - tuning_constant`

igvc/src/motor_controller/main.cpp
```diff
@@ -28,66 +35,164 @@ int main(int argc, char** argv)
...
+    for(int i = 0; i < 3; i++) {
+      std::string ret = port.readln();
+      if(!ret.empty() && ret.at(0) == '#' && ret.at(1) != 'E') {
+        size_t p_loc = ret.find('P');
+        size_t d_loc = ret.find('D');
+        size_t i_loc = ret.find('I');
+        size_t end = ret.find('\n');
+        if(p_loc != std::string::npos) {
+          std::vector<std::string> vals = split(ret.substr(p_loc + 1, end), ',');
+          ROS_INFO_STREAM("p ret = " << ret);
```
evan10s: This seems like a debug print statement.  I think we could either print nothing if everything is OK with the PID constants returned by the motor board or if there's an issue, print the expected vs. actual values sent back from the motor board

igvc/src/motor_controller/main.cpp
```diff
@@ -28,66 +35,164 @@ int main(int argc, char** argv)
...
+          if(vals.size() == 2) {
+            valid_values_i = true;
+            valid_values_i = (stod(vals.at(0)) == i_l) && valid_values_i;
+            valid_values_i = (stod(vals.at(1)) == i_r) && valid_values_i;
+          }
+        } else {
+          ROS_INFO_STREAM("recieved unknown string while setting PID values " << ret);
+        }
+      } else if(ret.empty()) {
+        ROS_ERROR_STREAM("Empty return from motor arduino while sending PID values.\t" << ret);
```
evan10s: Change `motor arduino` to `Motor Board`

igvc/src/motor_controller/main.cpp
```diff
@@ -28,66 +35,164 @@ int main(int argc, char** argv)
...
+      }
+    }
+    rate.sleep();
+  }
+  ROS_INFO_STREAM("sucessfully sent PID values");
```
evan10s: Sometimes the first letter of print statements are capitalized, but other time's it's not.  This is something that would ideally be consistent once we're out of the debugging stage.

igvc/launch/localization.launch
dpattison3: Good thought, but yes. The comment above this says to look above. It takes the same format as the one directly above, so the description is omitted for brevity.

evan10s: I see it now, great

igvc/launch/sensors.launch
JasonGibson274: I am rather sure that I is not currently implemented properly in firmware so that is why I is 0.

igvc/launch/imu.launch
barulicm: `output="screen"` is common for roslaunch nodes even when not debugging. Without this attribute, ROS logging macros (ie ROS_INFO_STREAM) write to a log file instead of the terminal.
See http://wiki.ros.org/roslaunch/XML/node#Attributes

JasonGibson274:
That was not compiling... you should not be satisfied.

JasonGibson274:
IMU calibration will become a separate PR

JasonGibson274:
IMU file will be completely rewritten for the new IMU so any IMU changes are moot and will be address in the next PR. Current changes include averaging out the last 1000 readings to account for intrinsic bias/tilt.

JasonGibson274:
@dpattison3 for older member final review and @atuttle7 for newer member review

JasonGibson274:
addressed in #323

PR #321, opened 2018-05-27
State estimator

dpattison3:
gtsam-based state estimator. Works in simulation. Worksish on real data (not a lot testing with good data).


JasonGibson274:
This can be reviewed as a separate thing but is in addressed in #323

PR #323, opened 2018-07-08
Jason master

JasonGibson274:
This is all of the code that was changed the month before competition and some during. I will be looking through this and removing the code that should not exist and/or was added during a hasty competition. This is being merged into devel and not master since I know for a fact there are open issue on this code that I think should be iterated on.

A short summary of changes

Path planner was converted to plan in discrete space rather than continuous space. 

A path follower was added to give smooth trajectories from the current location to a point a certain distance along the path. This separates path planning and following allowing the robot to drive longer without a new path. The paper it was adapted from is here tps://web.eecs.umich.edu/~kuipers/papers/Park-icra-11.pdf

The mapper was converted from continuous to discrete space using an occupancy grid. The occupancy grid is a cv::Matrix where the number represents the probability that a square is occupied. 

A new imu was added but the transform is wrong and requires debugging on the actual IMU. 

A new state estimator was added and has not worked on the actual robot yet. It requires tuning and we were too close to competition in order to do it justice. The state estimator is based on factor graphs and GTSAM.

I will be cleaning this PR up at time allows for me and making issues that will need to be resolved before this code should be merged into master.


igvc/launch/imu.launch
```diff
@@ -1,4 +1,21 @@
 <!-- imu.launch -->
 <launch>
-    <node name="imu" pkg="igvc" type="imu"/>
+    <node name="imu" pkg="igvc" type="imu" output="screen" >
+      <param name="yaw_offset" type="double" value="0"/>
```
JasonGibson274: This is due to gyro drift. Not much that can be done. There does not really seem to be a pattern.

JasonGibson274:
Okay I am going to fix the test cases for what is left here then we (@atuttle7 ) should do some testing on the actual robot to confirm that things are working then we can merge this PR.

PR #333, opened 2018-10-24
Fix point orientation operation in mapper node. Update simulation documentation

Alescontrela:
Fixed some stuff in the mapper node. Created an optional `plot_rviz` flag in mapper.launch that sets up an rviz visualization when `plot_rviz:=true`.

### DIFF
--- a/igvc_navigation/launch/localization.launch
+++ b/igvc_navigation/launch/localization.launch
@@ -21,7 +21,7 @@
 
       <!-- for if the IMU (after magnetic declination adjustment)
            does not read 0 -->
-      <param name="yaw_offset" value="1.57"/>
+      <param name="yaw_offset" value="0"/>
 
       <!-- this assumes we are in a planar environment -->
       <param name="two_d_mode" value="false"/>
--- a/igvc_navigation/launch/localization.launch
+++ b/igvc_navigation/launch/localization.launch
@@ -19,10 +19,6 @@
       <!-- does ardupilot account for magnetic declination? -->
       <param name="magnetic_declination_radians" value="0"/>
 
-      <!-- for if the IMU (after magnetic declination adjustment)
-           does not read 0 -->
-      <param name="yaw_offset" value="0"/>
-
       <!-- this assumes we are in a planar environment -->
       <param name="two_d_mode" value="false"/>
 
--- a/igvc_navigation/launch/pather.launch
+++ b/igvc_navigation/launch/pather.launch
@@ -3,7 +3,9 @@
 <!-- map.launch -->
 <launch>
     <!-- Waypoints -->
-    <!-- <include file="$(find igvc_navigation)/launch/waypoint_source.launch" /> -->
+    <arg name="simulation" default="false"/>
+    <include file="$(find igvc_navigation)/launch/waypoint_source.launch"
+             unless="$(arg simulation)"/>
 
     <!-- Planner -->
     <node name="path_planner" pkg="igvc_navigation" type="pathplanner" output="screen" >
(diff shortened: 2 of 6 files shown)


igvc_navigation/launch/localization.launch
```diff
@@ -21,7 +21,7 @@
 
       <!-- for if the IMU (after magnetic declination adjustment)
            does not read 0 -->
-      <param name="yaw_offset" value="1.57"/>
+      <param name="yaw_offset" value="0"/>
```
JasonGibson274: We should double check this on the robot. This may be a on the robot configuration. In that case we should rotate our IMU in the URDF to match whatever is on the robot. 

TLDR; check IMU orientation IRL and make sure simulation matches not the other way around.

igvc_navigation/launch/pather.launch
```diff
@@ -3,7 +3,7 @@
 <!-- map.launch -->
 <launch>
     <!-- Waypoints -->
-    <include file="$(find igvc_navigation)/launch/waypoint_source.launch" />
+    <!-- <include file="$(find igvc_navigation)/launch/waypoint_source.launch" /> -->
```
JasonGibson274: so this is a simulation thing that breaks IRL. If we want to restructure things so that this is not an issue now is the time for that.

igvc_rviz_plugins/configs/jessii_conf.rviz
```diff
@@ -0,0 +1,289 @@
+Panels:
```
JasonGibson274: We should have more of these but I am not sure how to best organize them. Opening up an issue for it

igvc_navigation/launch/localization.launch
Alescontrela: So I talked to justin about this yesterday. He mentioned we no longer need the 'yaw_offset' parameter. The reason it was needed at competition was because the IMU wasn't given enough time to converge on an accurate yaw value, so it set its initial yaw to an unfiltered, inaccurate value - this required manual correction via the yaw_offset parameter.

igvc_navigation/launch/pather.launch
Alescontrela: I agree. A simple workaround would be to include a launch file flag, `simulation`, that doesn't load IRL waypoints when set to true.

Let me know if you think this would be a good approach.

igvc_rviz_plugins/configs/jessii_conf.rviz
Alescontrela: So the reason i used this specific configuration is because it highlights everything the robot is capable of doing at the moment:
* sensing with the lidar
* creating an occupancy grid
* planning and following paths using the occupancy grid

If you want, i can remove it from this PR but I think it adds value as a simple debugging tool.

igvc_navigation/launch/localization.launch
JasonGibson274: Then I would remove the parameter entirely, but it was mostly used for actual offset.

igvc_rviz_plugins/configs/jessii_conf.rviz
JasonGibson274: No leave it in. I like it. I just mean later on we should organize them

igvc_navigation/launch/pather.launch
JasonGibson274: yes, let's go with that for now

igvc_navigation/launch/localization.launch
Alescontrela: o.k. i'll get rid of it. I'll still include it as a parameter in the documentation just for reference.

Alescontrela:
The new PR should include the above-mentioned changes 👍

PR #337, opened 2018-10-31
Refactored mapper file based on 10/28 meeting discussions

oswinso:
1. Created parameters `increment_step` and `occupancy_grid_threshold` in launch file to replace magic numbers.
2. Created helper functions `checkExistsStaticTransform`, `handlePointCloud` and `setMsgValues` to declutter `frame_callback`.
3. Changed occupancy grid to increase values even when the value is between 230 and 255.
4. Fixed check for missing parameters to properly check if **any** of the parameters are missing, instead of all the parameters being missing (De Morgan's Law: gotta change && to ||)

### DIFF
--- a/igvc_navigation/launch/mapper.launch
+++ b/igvc_navigation/launch/mapper.launch
@@ -8,6 +8,7 @@
             <param name="occupancy_grid_width" type="int" value="50" />
             <param name="occupancy_grid_resolution" type="double" value="0.2" />
             <param name="occupancy_grid_threshold" type="int" value="178" />
+            <param name="decay_period" type="double" value="0.1" />
             <param name="start_X" type="double" value="25"/>
             <param name="start_Y" type="double" value="25"/>
             <param name="increment_step" type="int" value="125"/>
--- a/igvc_navigation/launch/mapper.launch
+++ b/igvc_navigation/launch/mapper.launch
@@ -8,7 +8,8 @@
             <param name="occupancy_grid_width" type="int" value="50" />
             <param name="occupancy_grid_resolution" type="double" value="0.2" />
             <param name="occupancy_grid_threshold" type="int" value="178" />
-            <param name="decay_period" type="double" value="0.1" />
+            <param name="decay_period" type="double" value="0.5" />
+            <param name="transform_max_wait_time" type="double" value="3.0" />
             <param name="start_X" type="double" value="25"/>
             <param name="start_Y" type="double" value="25"/>
             <param name="increment_step" type="int" value="125"/>
--- a/igvc_navigation/launch/mapper.launch
+++ b/igvc_navigation/launch/mapper.launch
@@ -7,8 +7,10 @@
             <param name="occupancy_grid_length" type="int" value="50" />
             <param name="occupancy_grid_width" type="int" value="50" />
             <param name="occupancy_grid_resolution" type="double" value="0.2" />
+            <param name="occupancy_grid_threshold" type="int" value="178" />
             <param name="start_X" type="double" value="25"/>
             <param name="start_Y" type="double" value="25"/>
+            <param name="increment_step" type="int" value="125"/>
             <param name="debug" type="bool" value="true"/>
     </node>
 
--- a/igvc_navigation/launch/mapper.launch
+++ b/igvc_navigation/launch/mapper.launch
@@ -8,8 +8,8 @@
             <param name="occupancy_grid_width" type="int" value="50" />
             <param name="occupancy_grid_resolution" type="double" value="0.2" />
             <param name="occupancy_grid_threshold" type="int" value="178" />
-            <param name="decay_period" type="double" value="10000" />
-            <param name="transform_max_wait_time" type="double" value="3.0" />
+            <param name="decay_period" type="double" value="60" />
+            <param name="transform_max_wait_time" type="double" value="0.5" />
             <param name="start_X" type="double" value="25"/>
             <param name="start_Y" type="double" value="25"/>
             <param name="increment_step" type="int" value="125"/>
--- a/igvc_navigation/launch/mapper.launch
+++ b/igvc_navigation/launch/mapper.launch
@@ -8,6 +8,7 @@
             <param name="occupancy_grid_width" type="int" value="50" />
             <param name="occupancy_grid_resolution" type="double" value="0.2" />
             <param name="occupancy_grid_threshold" type="int" value="178" />
+            <!-- Set decay_period to 0 to disable decay -->
             <param name="decay_period" type="double" value="60" />
             <param name="transform_max_wait_time" type="double" value="0.5" />
             <param name="start_X" type="double" value="25"/>
--- a/igvc_navigation/src/mapper/main.cpp
+++ b/igvc_navigation/src/mapper/main.cpp
@@ -10,9 +10,9 @@
 
 cv_bridge::CvImage img_bridge;
 
-ros::Publisher map_pub;
-ros::Publisher debug_pub;
-ros::Publisher debug_pcl_pub;
+ros::Publisher map_pub;                  // Publishes map
+ros::Publisher debug_pub;                // Debug version of above
+ros::Publisher debug_pcl_pub;            // Publishes map as individual PCL points
 std::unique_ptr<cv::Mat> published_map;  // matrix will be publishing
 std::map<std::string, tf::StampedTransform> transforms;
 std::unique_ptr<tf::TransformListener> tf_listener;
--- a/igvc_navigation/src/mapper/main.cpp
+++ b/igvc_navigation/src/mapper/main.cpp
@@ -44,7 +44,10 @@ std::tuple<double, double> rotate(double x, double y)
   return (std::make_tuple(newX, newY));
 }
 
-// Updates state with latest tf transform using the timestamp of the pcl::PointCloud msg
+/**
+ * Updates <code>RobotState state</code> with the latest tf transform using the timestamp of the message passed in
+ * @param msg <code>pcl::PointCloud</code> message with the timestamp used for looking up the tf transform
+ */
 void getOdomTransform(const pcl::PointCloud<pcl::PointXYZ>::ConstPtr &msg)
 {
   tf::StampedTransform transform;
@@ -58,6 +61,13 @@ void getOdomTransform(const pcl::PointCloud<pcl::PointXYZ>::ConstPtr &msg)
 }
 
 // Populates igvc_msgs::map message with information from sensor_msgs::Image and the timestamp from pcl_stamp
+/**
+ * Populates <code>igvc_msgs::map message</code> with information from <code>sensor_msgs::Image</code> and the
+ * timestamp from <code>pcl_stamp</code>
+ * @param message message to be filled out
+ * @param image image containing map data to be put into <code>message</code>
+ * @param pcl_stamp time stamp from the pcl to be used for <code>message</code>
+ */
 void setMsgValues(igvc_msgs::map &message, sensor_msgs::Image &image, uint64_t pcl_stamp)
 {
   pcl_conversions::fromPCL(pcl_stamp, image.header.stamp);
@@ -74,7 +84,10 @@ void setMsgValues(igvc_msgs::map &message, sensor_msgs::Image &image, uint64_t p
   message.y_initial = start_y;
 }
 
-// Updates the occupancy grid using information from the pcl::PointCloud transformed
+/**
+ * Updates the occupancy grid using information from the <code>pcl::PointCloud transformed</code>
+ * @param transformed Reference to pointcloud containing information from lidar / segmentation.
+ */
 void updateOccupancyGrid(const pcl::PointCloud<pcl::PointXYZ>::Ptr &transformed)
 {
   int offMapCount = 0;
@@ -111,7 +124,11 @@ void updateOccupancyGrid(const pcl::PointCloud<pcl::PointXYZ>::Ptr &transformed)
   }
 }
 
-// Checks if transform from base_footprint to msg.header.frame_id exists
+/**
+ * Checks if transform from base_footprint to msg.header.frame_id exists
+ * @param msg
+ * @param topic
+ */
 void checkExistsStaticTransform(const pcl::PointCloud<pcl::PointXYZ>::ConstPtr &msg, const std::string &topic)
 {
   if (transforms.find(topic) == transforms.end())
@@ -133,7 +150,9 @@ void checkExistsStaticTransform(const pcl::PointCloud<pcl::PointXYZ>::ConstPtr &
   }
 }
 
-// Decays map by 1 universally on callback
+/**
+ * Decays map by 1 universally on callback
+ */
 void decayMap(const ros::TimerEvent &)
 {
   int nRows = published_map->rows;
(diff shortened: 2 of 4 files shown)


igvc_navigation/src/mapper/main.cpp
```diff
@@ -95,78 +136,46 @@ void frame_callback(const pcl::PointCloud<pcl::PointXYZ>::ConstPtr &msg, const s
...
-      }
-      count++;
-    }
-    else if (!offMap)
-    {
-      ROS_WARN_STREAM("Some points out of range, won't be put on map.");
-      offMap = true;
-    }
-  }
   img_bridge = cv_bridge::CvImage(msgBoi.header, sensor_msgs::image_encodings::MONO8, *published_map);
```
JasonGibson274: change to new image size

igvc_navigation/src/mapper/main.cpp
```diff
@@ -95,78 +136,46 @@ void frame_callback(const pcl::PointCloud<pcl::PointXYZ>::ConstPtr &msg, const s
...
   img_bridge.toImageMsg(imageBoi);  // from cv_bridge to sensor_msgs::Image
   time = ros::Time::now();          // so times are exact same
```
JasonGibson274: What are you trying to accomplish here?

igvc_navigation/src/mapper/main.cpp
```diff
@@ -95,78 +136,46 @@ void frame_callback(const pcl::PointCloud<pcl::PointXYZ>::ConstPtr &msg, const s
...
-  msgBoi.resolution = resolution;
-  msgBoi.orientation = orientation;
-  ROS_INFO_STREAM("robot location " << std::round(cur_x / resolution) + start_x << ", "
-                                    << std::round(cur_y / resolution) + start_y);
-  msgBoi.x = std::round(cur_x / resolution) + start_x;
-  msgBoi.y = std::round(cur_y / resolution) + start_y;
-  msgBoi.x_initial = start_x;
-  msgBoi.y_initial = start_y;
+  setMsgValues(time);
+  ROS_INFO_STREAM("robot location " << std::round(x_robot / resolution) + start_x << ", "
```
JasonGibson274: I'd cut this. It is just annoying. Or better yet print if debug only

igvc_navigation/src/mapper/main.cpp
```diff
@@ -95,78 +136,46 @@ void frame_callback(const pcl::PointCloud<pcl::PointXYZ>::ConstPtr &msg, const s
...
-    // ROS_INFO_STREAM("\nThe robot is located at " << cur_x << "," << cur_y << "," << orientation);
+    // ROS_INFO_STREAM("\nThe robot is located at " << x_robot << "," << y_robot << "," << orientation);
     pcl::PointCloud<pcl::PointXYZRGB>::Ptr fromOcuGrid =
         pcl::PointCloud<pcl::PointXYZRGB>::Ptr(new pcl::PointCloud<pcl::PointXYZRGB>());
     for (int i = 0; i < width_x; i++)
     {
       for (int j = 0; j < length_y; j++)
       {
-        if (published_map->at<uchar>(i, j) >= (uchar)178)
+        if (published_map->at<uchar>(i, j) >= (uchar)occupancy_grid_threshold)
```
JasonGibson274: I would cast occupancy_grid_threshold to uchar and then direct compare instead of doing this each time

oswinso:
Thanks for the feedback. I'm still trying to get used to C++, and this is really useful for me.

igvc_navigation/src/mapper/main.cpp
```diff
@@ -208,8 +218,10 @@ int main(int argc, char **argv)
   pNh.getParam("occupancy_grid_length", length_y);
   pNh.getParam("occupancy_grid_width", width_x);
   pNh.getParam("occupancy_grid_resolution", resolution);
+  pNh.getParam("occupancy_grid_threshold", occupancy_grid_threshold);
   pNh.getParam("start_X", cont_start_x);
   pNh.getParam("start_Y", cont_start_y);
+  pNh.getParam("increment_step", increment_step);
   pNh.getParam("orientation", orientation);
```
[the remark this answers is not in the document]
oswinso: Can I ask what you mean by when you say that it is set in the callback? Do you mean that this should be set in the callback? This variable used to be just a magic number, and I wasn't sure where would be the best place to define it.

igvc_navigation/src/mapper/main.cpp
```diff
@@ -54,8 +57,8 @@ void odom_callback(const nav_msgs::Odometry::ConstPtr &msg)
 {
   if (update)
```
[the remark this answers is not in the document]
oswinso: I'm a bit confused about what you mean here. Are you saying that I should be synchronizing the timestamps from the odom message and the point cloud message? That would require a much larger change to the logic of the file right?

igvc_navigation/src/mapper/main.cpp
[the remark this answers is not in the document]
JasonGibson274: It's not a magic number, it is set in the callback, line 69. In fact this should never run set to zero since that would assume that we are facing east without and input as to why that is.

igvc_navigation/src/mapper/main.cpp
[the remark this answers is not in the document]
JasonGibson274: Yes this will be a significant change and we will have to validate it to make sure it works like I think it does.

You should be using the position message that most closely matches the timestamp of the sensor message (i.e. lidar or camera). Fortunately TF will handle this for you. What you would do here is get the transform from sensor to /odom each time the callback is called and then use that transform to get the points in to the global reference frame. TF will do linear interpolation between the two most recent transforms relative to the time you are requesting it at. Which should be close enough in theory.

JasonGibson274:
So it depends on what you are trying to do. Use a smart pointer when you really need it to act like a pointer (since it manages memory for you).  Use a unique pointer when only a single variable, without assigning it, can be used to scope the variable properly. Use a share pointer when you will need to have multiple references to the object with different lifetimes. References are just syntactic sugar for const pointers, so if you want to pass an object by reference (i.e. not copy and be able to edit the object when passed into a method) use a reference whenever possible.

igvc_navigation/src/mapper/main.cpp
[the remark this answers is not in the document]
JasonGibson274: Realistically we should implement the TF method and compare that to the current method and see which gives less error

igvc_navigation/src/mapper/main.cpp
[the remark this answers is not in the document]
oswinso: Oh, I thought you were referring to `increment_step`, which was why I was confused. You are referring to `orientation` here right?

igvc_navigation/src/mapper/main.cpp
oswinso: I think this converts published_map from CvImage to a sensor_msgs::Image so that we can publish it?

igvc_navigation/src/mapper/main.cpp
oswinso: Can you clarify what you mean by new image size?

igvc_navigation/src/mapper/main.cpp
JasonGibson274: I mean the time stamp specifically

igvc_navigation/src/mapper/main.cpp
JasonGibson274: CV16U vs CV8U, i.e. if we get a byte or two bytes to represent probability.

igvc_navigation/src/mapper/main.cpp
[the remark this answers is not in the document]
JasonGibson274: Nope, the bottom line is the line I am commenting on

igvc_navigation/src/mapper/main.cpp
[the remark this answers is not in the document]
oswinso: This would be done using `sync_policies` right? Do you if exact time or approximate time works better here?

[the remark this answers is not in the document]
oswinso: I'm actually a bit confused about where `odometry/filtered` is being published from, since I can't really find it. What's sending messages in that topic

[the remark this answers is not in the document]
JasonGibson274: EKF localization, a node we include and run (localization.launch) but we did not write ourselves. It is just the best guess of the current position.

igvc_navigation/src/mapper/main.cpp
oswinso: So are we switching to CV16U instead of CV8U?

igvc_navigation/src/mapper/main.cpp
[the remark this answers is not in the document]
oswinso: I'm still quite confused about what I need to do. My guess is that I do `lookupTransform` on the transform published by the localization node (not sure what the name of this is), and then I use the results of that transform rather than the messages I am getting from `/odometry/filtered`?

igvc_navigation/src/mapper/main.cpp
```diff
@@ -13,221 +15,281 @@
...
-        published_map->at<uchar>(point_x, point_y) += (uchar)125;
-      }
-      count++;
+        nCols *= nRows;
+        nRows = 1;
     }
-    else if (!offMap)
+    int i, j;
+    uchar* p;
+    for (i = 0; i < nRows; i++)
```
Codetector1374: Unrelated to code exactly. But I think we should have discussion on how we should handle the decay.

igvc_navigation/src/mapper/main.cpp
```diff
@@ -13,221 +15,281 @@
...
-tf::TransformListener *tf_listener;
-std::string topics;
-Eigen::Map<Eigen::Matrix<unsigned char, Eigen::Dynamic, Eigen::Dynamic>> *eigenRep;
+std::unique_ptr<tf::TransformListener> tf_listener;
 
 double resolution;
 double orientation;
 double _roll;
 double _pitch;
+double transform_max_wait_time;
```
Codetector1374: I don't remember we have decided on a naming scheme for global vars. But it seems like that some vars are already named starting with '_' for member variable. maybe we should address that change as well?

igvc_navigation/launch/mapper.launch
```diff
@@ -7,8 +7,12 @@
             <param name="occupancy_grid_length" type="int" value="50" />
             <param name="occupancy_grid_width" type="int" value="50" />
             <param name="occupancy_grid_resolution" type="double" value="0.2" />
+            <param name="occupancy_grid_threshold" type="int" value="178" />
+            <param name="decay_period" type="double" value="0.5" />
```
Alescontrela: This parameter should be configurable to allow for no decay at all. Consider a way to declare `decay_period` such that the map doesn't decay at all (perhaps a negative value).

igvc_navigation/src/mapper/main.cpp
Alescontrela: I think a recent commit to devel has changed independent roll, pitch, cur_x, cur_y into a `RobotState` object, for consistency try to incorporate that here.

igvc_navigation/src/mapper/main.cpp
```diff
@@ -13,221 +15,281 @@
...
 
-void odom_callback(const nav_msgs::Odometry::ConstPtr &msg)
-{
-  if (update)
-  {
-    cur_x = msg->pose.pose.position.x;
-    cur_y = msg->pose.pose.position.y;
-    tf::Quaternion quat;
-    tf::quaternionMsgToTF(msg->pose.pose.orientation, quat);
+//void odom_callback(const nav_msgs::Odometry::ConstPtr &msg)
```
Alescontrela: remove this commented chunk of code

igvc_navigation/src/mapper/main.cpp
Alescontrela: Same thought as @Codetector1374

igvc_navigation/src/mapper/main.cpp
```diff
@@ -13,221 +15,281 @@
...
+
+    std::list<ros::Subscriber> subs;
+    tf_listener = std::unique_ptr<tf::TransformListener>(new tf::TransformListener());
+
+    double cont_start_x;
+    double cont_start_y;
+    double decay_period;
+    int cont_occupancy_grid_threshold;
+
+    if (!(pNh.hasParam("topics") && pNh.hasParam("occupancy_grid_width") && pNh.hasParam("occupancy_grid_length") &&
```
Alescontrela: this would be the proper place to implement `NodeUtils::param()` or `NodeUtils::getParam()`

oswinso: It's a separate PR, so I intended for the change to happen during merging, which is why I didn't use the `igvc::getParam()` and `igvc::param()` here.

igvc_navigation/src/mapper/main.cpp
oswinso: @JasonGibson274 I think we've come to a conclusion on slack that for now decaying globally is fine right?

oswinso:
@Alescontrela I've fixed the merge conflict, and added checking for disabling decay.

igvc_navigation/src/mapper/main.cpp
```diff
@@ -45,105 +35,158 @@ std::tuple<double, double> rotate(double x, double y)
...
-      transform.frame_id_ = "/lidar";
-      transform.stamp_ = ros::Time::now();
-      transforms.insert(std::pair<std::string, tf::StampedTransform>(topic, transform));
     }
   }
-  pcl_ros::transformPointCloud(*msg, *transformed, transforms.at(topic));
+}
 
-  pcl::PointCloud<pcl::PointXYZ>::const_iterator point;
+void decayMap(const ros::TimerEvent &)
```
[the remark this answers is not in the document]
oswinso: what do you mean?

igvc_navigation/src/mapper/main.cpp
```diff
@@ -45,105 +35,158 @@ std::tuple<double, double> rotate(double x, double y)
   return (std::make_tuple(newX, newY));
 }
 
-void odom_callback(const nav_msgs::Odometry::ConstPtr &msg)
+void getOdomTransform(const pcl::PointCloud<pcl::PointXYZ>::ConstPtr &msg)
```
jzheng84: Can you add doxygen comments? Same with below, all of these functions don't have high level comments explaining what is actually happening.

igvc_navigation/src/mapper/main.cpp
```diff
@@ -155,24 +198,28 @@ void frame_callback(const pcl::PointCloud<pcl::PointXYZ>::ConstPtr &msg, const s
     fromOcuGrid->header.stamp = msg->header.stamp;
     debug_pcl_pub.publish(fromOcuGrid);
   }
-  update = true;
+  //  update = true;
```
jzheng84: delete

igvc_navigation/src/mapper/main.cpp
```diff
@@ -1,41 +1,31 @@
...
-#include <opencv2/core/eigen.hpp>
-#include <opencv2/opencv.hpp>
-#include "tf/transform_datatypes.h"
 #include <igvc_utils/RobotState.hpp>
 
-igvc_msgs::map msgBoi;  // >> message to be sent
 cv_bridge::CvImage img_bridge;
-sensor_msgs::Image imageBoi;  // >> image in the message
 
 ros::Publisher map_pub;
```
jzheng84: comment what each of these publishers are for, it's not immediately clear what the difference is

igvc_navigation/src/mapper/main.cpp
oswinso: fixed

oswinso:
the stack.nl site is down so I'm not too sure about how the autolinking etc stuff works

PR #340, opened 2018-11-12
Created wrapper for getParam and param

oswinso:
This PR creates a wrapper around pNh.getParam and pNh.param. The wrapper function will log an error then exit if getParam or param returns false.


Alescontrela:
This functionality might be better located in a more general `NodeUtils` class since I doubt the `ParamUtils` class will be expanded much. A `NodeUtils` class could contain all helper methods to be used during the lifetime of a node.

Also, make sure you update wiki to reflect any changes you've made.

oswinso:
Yeah, I think `NodeUtils` would make more sense.

Alescontrela:
http://wiki.ros.org/roscpp_tutorials/Tutorials/Parameters#getParam.28.29

igvc_perception/CMakeLists.txt
```diff
@@ -105,9 +106,9 @@ catkin_package(
 ## Your package locations should be listed before other locations
 # include_directories(include)
 include_directories(
-  ${catkin_INCLUDE_DIRS}
-  ${PCL_INCLUDE_DIRS}
-  include
+   ${catkin_INCLUDE_DIRS}
```
JasonGibson274: how many spaces is this?

igvc_utils/include/igvc_utils/NodeUtils.hpp
```diff
@@ -0,0 +1,28 @@
+#ifndef NODEUTILS_HPP
+#define NODEUTILS_HPP
+
+#include <ros/ros.h>
+#include <stdlib.h>
+
+namespace igvc
+{
+    template<class T>
+    void param(const ros::NodeHandle pNh, const std::string &param_name, T &param_val, const T &default_val)
```
JasonGibson274: So I just realized that this will not tell you from what node you came from. We should probably add that.

oswinso: You're referring to the messages that error messages that we get if the param isn't found right?

oswinso: I've added pNh.getNamespace(), which should be enough information to find out which node called the function I think.

oswinso:
I don't think it will be troublesome to use this wrapper class, since the arguments are almost exactly the same, but we get the added parameter checking basically for free. I intended this to replace `param()` and `getParam()`, and I do think I've changed all existing uses of `param()` and `getParam()` to use the equivalent wrapper functions.

igvc_utils/include/igvc_utils/NodeUtils.hpp
jzheng84: const ros::NodeHandle &pNh

JasonGibson274:
Run the clang formatter on this before merging though

PR #341, opened 2018-11-27
Fix egocentric heading calculations in Smooth Control law implementation

Alescontrela:
This commit includes Jason's preliminary updates to the path planner as well as my updates to the Smooth control law. Smooth Control Law implementation updates include:

- Fix how egocentric polar angle for delta is calculated

- Fix how egocentric polar angle for theta is calculated

- Add visualization for updated smooth control law implementation (still some TODOs)

Details regarding our smooth control law can be found [here](https://web.eecs.umich.edu/~kuipers/papers/Park-icra-11.pdf).


oswinso:
Do the fixes to the egocentric polar angles fix the path following problems we were having?

Alescontrela:
> Do the fixes to the egocentric polar angles fix the path following problems we were having?

The changes fix many of the errors present in the previous implementation. There are still some more subtle issues that need sorting.

igvc_navigation/launch/mapper.launch
```diff
@@ -4,8 +4,8 @@
     <!-- Map Node -->
     <node name="mapper" pkg="igvc_navigation" type="mapper" output="screen">
             <param name="topics" type="string" value="/scan/pointcloud /semantic_segmentation_cloud"/>
-            <param name="occupancy_grid_length" type="int" value="50" />
-            <param name="occupancy_grid_width" type="int" value="50" />
+            <param name="occupancy_grid_length" type="int" value="200" />
```
JasonGibson274: this is a YUGE matrix. Why?

If our course in sim is this big then make an issue for that.

igvc_navigation/launch/pather.launch
```diff
@@ -18,16 +18,17 @@
...
 
     <!-- Follower -->
     <node name="path_follower" pkg="igvc_navigation" type="path_follower" output="screen" >
         <param name="maximum_vel" type="double" value="6.0"/>
         <param name="target_v" type="double" value="0.6"/>
         <param name="axle_length" type="double" value="0.52"/>
-        <param name="lookahead_dist" type="double" value="1.8"/>
-        <param name="k1" type="double" value="0.5"/>
-        <param name="k2" type="double" value="1.0"/>
+        <param name="lookahead_dist" type="double" value="1.3"/>
```
JasonGibson274: might have to retune on the actual robot

igvc_navigation/src/path_follower/main.cpp
```diff
@@ -48,52 +71,82 @@ void position_callback(const nav_msgs::OdometryConstPtr& msg)
...
 
+  /**
+  Find the furthest point along trajectory that isn't further than the
+  lookahead distance. This is the target position.
+  */
   if (get_distance(cur_x, cur_y, end.x, end.y) > lookahead_dist)
   {
-    double distance = 0;
+    double distance = 0; // Eigen::Vector3d init_orientation = los - heading;
+  // cur_theta = orientation - std::atan2(init_orientation[1], init_orientation[0]);
```
JasonGibson274: is this supposed to be like this?

igvc_navigation/src/path_follower/main.cpp
```diff
@@ -22,17 +22,40 @@ SmoothControl controller;
 
 void path_callback(const nav_msgs::PathConstPtr& msg)
 {
-  ROS_INFO("Follower got path");
+  ROS_INFO_STREAM("Follower got path. Size: " << msg->poses.size());
   path = msg;
 }
 
 double get_distance(double x1, double y1, double x2, double y2)
```
oswinso: I think we could probably extract this out of this class and into the robot_state class maybe?

igvc_navigation/src/path_follower/main.cpp
```diff
@@ -22,17 +22,40 @@ SmoothControl controller;
...
 {
+  /**
+  Calculates euclidian distance between two points
+  */
   return sqrt(pow(x2 - x1, 2) + pow(y2 - y1, 2));
 }
 
+void compute_angle(float& angle, Eigen::Vector3d vec2, Eigen::Vector3d vec1)
```
oswinso: This one I think we could also extract out of here and into robot_state to make this class cleaner

igvc_navigation/src/path_planner/GraphSearch.hpp
```diff
@@ -1,17 +1,46 @@
...
 #include <stack>
+#include <map> // TODO move to unordered map
 
 #include "SearchProblem.hpp"
 
+// TODO this is kind of an issue
 using namespace std;
 
+template <class StateType, class ActionType>
+class StateComparator {
```
oswinso: This stuff here is unrelated to the Smooth Control stuff right? I think it belongs in another PR so we can keep the PRs focused? @JasonGibson274

igvc_navigation/launch/mapper.launch
Alescontrela: The robot traveled too far to stay within the initial occupancy grid length. I increased the size so that I could see how it performs on the rest of the course. 200m is indeed pretty big, i'll reduce the value of this parameter on my next commit. 

What do you mean by comments inside the method? Like method explanations? i.e.

```c++

int foo(int a) 
{
    /**
    does bar on foo 
    */
    return 0;
}
```

JasonGibson274: yes

igvc_navigation/launch/pather.launch
Alescontrela: this is an important thing we have to consider. The paper doesn't really do the best job explaining how these parameters affect the trajectory but, from observing the robot's behavior with different values of k1 and k2 I noticed that k1 affects how theta changes relative to the distance from the target pose and k2 affects how quickly the robot converges to the target pose.

I found these values of k1 and k2 to do a pretty good job at getting the robot on the path

igvc_navigation/src/path_follower/main.cpp
Alescontrela: no. I'll fix it

igvc_navigation/src/path_follower/main.cpp
Alescontrela: I'm not sure what you mean by this. Could you clarify a bit further?

igvc_navigation/src/path_follower/main.cpp
Alescontrela: This isn't a class. 

This helper method could definitely be placed in a separate class, although i'm not sure whether it should be put in `RobotState` or `NodeUtils`.

igvc_navigation/launch/mapper.launch
Alescontrela: O.k. I think your approach is the correct one. I'll fix this.

igvc_navigation/src/path_follower/main.cpp
JasonGibson274: as is node utils, then make a RobotState method that gives it to you for the states

igvc_navigation/src/path_follower/main.cpp
JasonGibson274: I think he means using robot state here, then have a method called distance that does this in that implementation.

igvc_navigation/src/path_planner/GraphSearch.hpp
JasonGibson274: ideally yes, but it is probably fine as is. this all goes into devel and that can be kind of chaotic.

igvc_navigation/src/path_follower/main.cpp
oswinso: I'm not too sure what you mean by "as in node utils", could you explain what you mean?

igvc_navigation/src/path_follower/SmoothControl.cpp
```diff
@@ -35,81 +31,57 @@ void SmoothControl::getTrajectory(igvc_msgs::velocity_pair& vel, nav_msgs::Path&
...
     pose.pose.position.y = end_pos[1];
     trajectory.poses.push_back(pose);
-    // std::cout << std::endl;
     cur_pos = end_pos;
   }
-  // std::cout << "\n end" << std::endl;
 }
 
-void SmoothControl::getAction(Eigen::Vector3d& result, Eigen::Vector3d cur_pos, Eigen::Vector3d target)
+void SmoothControl::getAction(Eigen::Vector3d& result, Eigen::Vector3d cur_pos,
```
AaronMcDaniel: Small detail, but is there a reason why we don't combine position and heading into one 5d vector?
It is a little confusing as to which position vector it corresponds to (I'm assuming it is the current position).

igvc_navigation/src/path_follower/SmoothControl.cpp
```diff
@@ -35,81 +31,57 @@ void SmoothControl::getTrajectory(igvc_msgs::velocity_pair& vel, nav_msgs::Path&
...
-  // std::cout << "theta = " << theta << " delta = " << delta << std::endl;
+  // adjust both angles to lie between -PI and PI
+  fitToPolar(delta);
+  fitToPolar(theta);
 
+  // calculate the radius of curvature, K
   double d = sqrt(pow(target[0] - cur_pos[0], 2) + pow(target[1] - cur_pos[1], 2));
-
-  double K = k2 * (delta - atan(k1 * theta));
+  double K = k2 * (delta - atan(-k1 * theta));
```
AaronMcDaniel: Did you mean to negate the theta? This should flip the quadrants we are operating in, but I'm not sure why that is necessary.

igvc_navigation/src/path_follower/SmoothControl.h
```diff
@@ -1,3 +1,12 @@
+/**
+Implementation of smooth control law derived from:
+A Smooth Control Law for Graceful Motion of Differential Wheeled Mobile Robots in 2D Environment
+
+Paper found here:
+
+https://web.eecs.umich.edu/~kuipers/papers/Park-icra-11.pdf
```
AaronMcDaniel: Much thanks for the citation

igvc_navigation/src/path_follower/main.cpp
AaronMcDaniel: He means the way it is now it should be put in NodeUtils. I agree since it is a general operation that could be used by a lot of nodes, not just the ones relating to state

JasonGibson274: as is, put it in node utils.

igvc_navigation/src/path_planner/igvcsearchproblem.cpp
```diff
@@ -1,117 +1,190 @@
...
+  if (x < 0 || y < 0 || x >= Map->image.size().width || y >= Map->image.size().height)
+  {
+    return false;
+  }
+  double min_val;
+  double max_val;
+  double sep = CSpace / Resolution;
+
+  // TODO how does this function worik
+  // implicit cast, TODO check a thing we want to do
```
AaronMcDaniel: Did you turn these TODOs into TODONEs? If so, then you may want to add a comment or two describing what this does. If not, then can you explain what these are?

igvc_rviz_plugins/configs/jessii_conf.rviz
```diff
@@ -272,10 +295,10 @@ Visualization Manager:
...
...fb000000120056006900650077007300200054006f006f02000001df000002110000018500000122fb000000200054006f006f006c002000500072006f0070006500720074006900650073003203000002880000011d000002210000017afb000000100044006900730070006c006100790073010000006e000004cf0000018200fffffffb0000002000730065006c0065006300740069006f006e00200062007500660066006500720200000138000000aa0000023a00000294fb00000014005700690064006500530074006500720065006f02000000e6000000d2000003ee0000030bfb0000000c004b0069006e0065006300740200000186000001060000030c00000261000000010000015f000004cffc0200000003fb0000001e0054006f006f006c002000500072006f00700065007200740069006500730100000041000000780000000000000000fb0000000a00560069006500770073010000006e000004cf0000013200fffffffb0000001200530065006c0065006300740069006f006e010000025a000000b200000000000000000000000200000490000000a9fc0100000001fb0000000a00560069006500770073030000004e00000080000002e100000197000000030000097a0000005afc0100000002fb0000000800540069006d006501000000000000097a0000057100fffffffb0000000800540069006d006501000000000000045000000000000000000000060c000004cf00000004000000040000000800000008fc0000000100000002000000010000000a0054006f006f006c00730100000000ffffffff0000000000000000
```
AaronMcDaniel: I am not sure if I'm comfortable with your change to bit 217, could you explain that change?

igvc_navigation/src/path_follower/main.cpp
Alescontrela: Ok gotcha. I'll create a `RobotState` object to hold odometry information and create an instance method that returns the euclidian distance to an arbitrary point.

igvc_navigation/src/path_follower/SmoothControl.cpp
Alescontrela: so cur_pos (current pose, _not_ current position) and egocentric_heading are a bit different. cur_pose holds odometry information relative to odom, (x,y,yaw). egocentric_heading holds two angles that are essential to calculating a trajectory but are not relative to odom, they're instead relative to the line of sight from the robot to the target pose.

If you would like a more detailed explanation, the paper does a good job explaining what these parameters mean

igvc_navigation/src/path_follower/SmoothControl.cpp
Alescontrela: I'm fairly certain this is the correct implementation. I think this has to do with the fact that delta and theta are egocentric (turn left is + , right is -), hence the negation.

igvc_navigation/src/path_planner/igvcsearchproblem.cpp
Alescontrela: @JasonGibson274

igvc_navigation/src/path_follower/SmoothControl.cpp
AaronMcDaniel: That makes sense. Thanks for clearing that up

igvc_navigation/src/path_follower/main.cpp
```diff
@@ -15,22 +17,36 @@ ros::Publisher target_pub;
...
 
-double lookahead_dist, maximum_vel;
+double stop_dist, maximum_vel;
 
 SmoothControl controller;
 
 void path_callback(const nav_msgs::PathConstPtr& msg)
 {
-  ROS_INFO("Follower got path");
+  ROS_INFO_STREAM("Follower got path. Size: " << msg->poses.size());
```
JasonGibson274: put under debug

igvc_navigation/src/path_planner/GraphSearch.hpp
```diff
@@ -1,17 +1,46 @@
...
 #include <list>
 #include <queue>
 #include <set>
 #include <stack>
+#include <map> // TODO move to unordered map
 
 #include "SearchProblem.hpp"
 
+// TODO this is kind of an issue
 using namespace std;
```
atuttle7: This should be removed.

igvc_navigation/src/path_planner/GraphSearch.hpp
```diff
@@ -67,121 +96,73 @@ struct FrontierTraits<StateType, ActionType, std::queue>
...
+    std::map<StateType, std::pair<StateType, ActionType>> predecessorList;
+
+    // if no maximum iterations are given go int max
+    if(maxIter == 0) {
+      maxIter = std::numeric_limits<int>::max();
     }
 
+    frontier.push(problem.getStartState());
+
     auto iteration = 0;
```
atuttle7: Prefer anything but auto, if possible

igvc_navigation/src/path_planner/SearchProblem.hpp
```diff
@@ -28,6 +30,22 @@ class Path
   {
     actions.push_back(action);
   }
+  void addStateActionRev(StateType state, ActionType action) {
```
atuttle7: newline

igvc_navigation/src/path_planner/SearchProblem.hpp
```diff
@@ -28,6 +30,22 @@ class Path
...
+    states.push_front(state);
+    actions.push_front(action);
+  }
+  void addStateActionPair(StateType state, ActionType action, double cost) {
```
atuttle7: newline

igvc_navigation/src/path_planner/SearchProblem.hpp
```diff
@@ -28,6 +30,22 @@ class Path
...
+    pathCost += cost;
+    states.push_back(state);
+    actions.push_back(action);
+  }
+  double getPathCost() const {
```
atuttle7: newline

igvc_navigation/src/path_planner/SearchProblem.hpp
```diff
@@ -28,6 +30,22 @@ class Path
...
+    return pathCost;
+  }
+  void setPathCost(double pathCost) {
```
atuttle7: linenew

igvc_navigation/src/path_follower/main.cpp
```diff
@@ -48,113 +64,78 @@ void position_callback(const nav_msgs::OdometryConstPtr& msg)
     return;
   }
 
+  double end_x = waypoint->point.x;
```
JasonGibson274: I would in line these with the function call for clarity, i.e. use waypoint->point.x when you get the distance

igvc_navigation/src/path_planner/main.cpp
```diff
@@ -1,5 +1,6 @@
-// convolve over the map
-// update the map based on incremental updates and update the convolve
+/*
+ * TODO explain
```
atuttle7: please address todo, and add general documentation to file

igvc_navigation/src/path_follower/main.cpp
```diff
@@ -48,113 +64,78 @@ void position_callback(const nav_msgs::OdometryConstPtr& msg)
...
-    tar_theta = yDiff > 0 ? M_PI : -M_PI;
-  }
-  else
-  {
-    tar_theta = atan2((yDiff), (xDiff));
-  }
+  double goal_dist = get_distance(cur_x, cur_y, end_x, end_y);
+  ROS_INFO_STREAM("Distance to waypoint: " << goal_dist << "(m.)");
 
   ros::Time time = ros::Time::now();
```
JasonGibson274: use message time

igvc_navigation/src/path_planner/main.cpp
```diff
@@ -93,41 +110,53 @@ int main(int argc, char** argv)
...
-  ros::Rate rate(40.0);
+  ros::Rate rate(rateTime);
   while (ros::ok())
   {
     ros::spinOnce();
 
-    auto distance_to_goal = search_problem.Start.distTo(search_problem.Goal, search_problem.Resolution);
-    if (!received_waypoint || distance_to_goal == 0 || distance_to_goal > 60)
+    // do not plan if you do not have a waypoint
+    if (!received_waypoint) {
```
atuttle7: /n

igvc_navigation/src/path_planner/main.cpp
```diff
@@ -93,41 +110,53 @@ int main(int argc, char** argv)
...
     search_problem.DistanceToGoal = search_problem.Start.distTo(search_problem.Goal, search_problem.Resolution);
-    path = GraphSearch::AStar(search_problem, expanded_callback);
+    path = GraphSearch::AStar(search_problem, expanded_callback, maxIter);
     nav_msgs::Path path_msg;
+    // TODO timestamp why
     path_msg.header.stamp = ros::Time::now();
     path_msg.header.frame_id = "odom";
     for (auto loc : *(path.getStates()))
-    {
+      {// publish path with theta for cool viz
```
atuttle7: move comment out of the bracket's line

igvc_navigation/src/path_follower/main.cpp
```diff
@@ -48,113 +64,78 @@ void position_callback(const nav_msgs::OdometryConstPtr& msg)
...
+      nav_msgs::Path trajectory_msg;
+      trajectory_msg.header.stamp = time;
+      trajectory_msg.header.frame_id = "/odom";
+
+      controller.getTrajectory(vel, path, trajectory_msg, cur_pos);
+      // publish trajectory
+      trajectory_pub.publish(trajectory_msg);
+
+      // publish target position
+      double tar_x = controller.target[0], tar_y = controller.target[1];
```
JasonGibson274: I really don't like this on a style side

igvc_navigation/src/path_follower/main.cpp
```diff
@@ -48,113 +64,78 @@ void position_callback(const nav_msgs::OdometryConstPtr& msg)
...
+      geometry_msgs::PointStamped target_point;
+      target_point.header.frame_id = "/odom";
+      target_point.header.stamp = time;
+      target_point.point.x = tar_x;
+      target_point.point.y = tar_y;
+      target_pub.publish(target_point);
+
+      ROS_INFO_STREAM("Distance to target: "
```
JasonGibson274: debug

igvc_navigation/src/path_planner/igvcsearchproblem.cpp
```diff
@@ -1,117 +1,190 @@
...
-    {
-      acts.push_back(move);
-    }
-    move = SearchMove(0, -TurningSpeed, deltat);
-    if (Octree->getTreeDepth() == 0 || isActionValid(move, state))
-    {
-      acts.push_back(move);
-    }
+  double theta = state.Theta;
+  // TODO this should not be needed
```
atuttle7: Why? Please explain

igvc_navigation/src/path_planner/igvcsearchproblem.cpp
```diff
@@ -1,117 +1,190 @@
...
+  double thetaThreshold = 0.1;
+  // TODO is this an issue
```
atuttle7: Why? Please explain

igvc_navigation/src/path_planner/igvcsearchproblem.cpp
```diff
@@ -1,117 +1,190 @@
...
+    // TODO handle case
+    //std::cerr << "\n\n\n\n\nfail" << action << "\n\n\n\n\n" << std::endl;
   }
+  result.PrevTheta.resize(ThetaChangeWindow);
+  double thetaDiff;
+  if (std::abs(state.Theta - M_PI) < 0.1)
+  {
+    thetaDiff = result.Theta > 0 ? result.Theta - M_PI : result.Theta + M_PI;
+  }
+  else if (std::abs(result.Theta - M_PI) < 0.1)
```
atuttle7: Could you explain with a comment/nice variable name what these .1's are?

igvc_navigation/src/path_follower/SmoothControl.h
```diff
@@ -8,21 +17,39 @@
...
+  Eigen::Vector3d cur_pos;
+  Eigen::Vector3d target;
+
+  void getTrajectory(igvc_msgs::velocity_pair& vel, nav_msgs::PathConstPtr path,
+        nav_msgs::Path& trajectory, Eigen::Vector3d cur_pos);
 
 private:
-  void getAction(Eigen::Vector3d& result, Eigen::Vector3d cur_pos, Eigen::Vector3d target);
-  void getResult(Eigen::Vector3d& result, Eigen::Vector3d cur_pos, Eigen::Vector3d action);
+  Eigen::Vector3d _los;
```
JasonGibson274: check the actual style guide

igvc_navigation/src/path_follower/SmoothControl.cpp
```diff
@@ -1,137 +1,320 @@
...
+    this->cur_pos << cur_pos;
+
+    // TODO: Create initialize_variables() method
+    unsigned int path_index;
+    path_index = getClosestPosition(path);
+    getTargetPosition(path, path_index);
+    getLineOfSightAndHeading();
+    getEgocentricAngles(path, path_index);
+
+    double dt = rollOutTime/10;
```
JasonGibson274: make a update rate parameter

igvc_navigation/src/path_follower/SmoothControl.cpp
```diff
@@ -1,137 +1,320 @@
...
+
+    for (int i = 0; i < 15; i++)
```
JasonGibson274: magic number, works in practice

igvc_navigation/src/path_planner/GraphSearch.hpp
```diff
@@ -67,121 +96,73 @@ struct FrontierTraits<StateType, ActionType, std::queue>
...
+          Path<StateType, ActionType> path;
+          reconstructPath(path, predecessorList, last);
           return path;
         }
-        list<ActionType> legalActions = problem.getActions(last, problem.getStartState());
+        list<ActionType> legalActions = problem.getActions(last);
 
-        for (typename list<ActionType>::iterator it = legalActions.begin(); it != legalActions.end(); it++)
-        {
+        for (typename list<ActionType>::iterator it = legalActions.begin();
```
kkeirstead3: list should be std::list (shouldn't be relying on std namespace)

PR #345, opened 2019-01-24
D* Lite and Field D*

Alescontrela:
This PR contains implementations for the D* Lite and Field D* incremental path planning algorithms. D* Lite is pretty much wrapped up, meaning it returns optimal paths in real time and has logic to account for edge cases. Field D* still has some bugs that need working out, which is the main motive behind this PR. Please let me know if you spot a logic error or some syntax error that might be an issue.

**What works:**
* D* Lite
* Field D* initial path planning capabilities (albeit slow)

**What doesn't work:**
1. Field D* cost map updates (when an obstacle is inserted, g-values should update to infinity but they don't)
2. As a result of 1, Field D* re-planning capabilities are inconsistent.

**Why it doesn't work:**
When a cell undergoes edge cost updates, all nodes surrounding the cell (and its C-space) are updated, meaning the affected nodes' RHS values are re-calculated and are re-inserted into the priority queue with a key. Nodes are then expanded from the priority queue based on their key value (low to high)  until the key of the top node in the PQ is greater than or equal to the start node's key and the start node is locally consistent (g == rhs). When nodes around updated cells are re-inserted into the priority queue with updated RHS values, their key values are higher than the key value of the start cell, meaning they are not expanded and the cost map is not updated to reflect new obstacles. 

**How to run:**
```bash
$ roslaunch igvc_navigation field_d_planner.launch simulation:=true # terminal window 3
```
**References**:
[D* Lite Paper](http://idm-lab.org/bib/abstracts/papers/aaai02b.pdf)
[Field D* Paper](https://www.ri.cmu.edu/pub_files/pub4/ferguson_david_2005_3/ferguson_david_2005_3.pdf)


Alescontrela:
Closing pull request to create individual PRs for D* Lite and Field D*

PR #349, opened 2019-02-04
D* Lite Path Planning Implementation

Alescontrela:
This PR contains the implementation for the D* Lite Incremental Path Planning Algorithm. The algorithm plans/replans optimal paths (with pi/4) increments in real time. 

**Usage**:
```bash
$ roslaunch igvc_navigation d_lite_planner.launch simulation:=true # terminal window 3
```
**References**:
[D* Lite](http://idm-lab.org/bib/abstracts/papers/aaai02b.pdf)
[Optimal and Efficient Path Planning for Unknown and Dynamic Environments (D*)](https://pdfs.semanticscholar.org/77e9/b970024bc5da2b726491823f7d617a303811.pdf)
[MIT Advanced Lecture 1: Incremental Path Planning](https://ocw.mit.edu/courses/aeronautics-and-astronautics/16-412j-cognitive-robotics-spring-2016/videos-for-advanced-lectures/advanced-lecture-1/)


igvc_navigation/src/d_lite_planner/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+add_executable(d_lite_planner
+        ${PROJECT_SOURCE_DIR}/include/${PROJECT_NAME}/Graph.cpp
```
barulicm: I'm confused by the file organization for this target. Why are all of these files in `include`?

igvc_navigation/src/d_lite_planner/main.cpp
```diff
@@ -0,0 +1,245 @@
...
+#include <tf/transform_datatypes.h>
+
+#include <mutex>
+#include <tuple>
+#include <limits>
+#include <vector>
+#include "igvc_navigation/Graph.h"
+#include "DLitePlanner.h"
+
+std::mutex planning_mutex;
```
barulicm: There are a lot of pretty heavy objects in the global scope here. I'd suggest spending some time to figure out if this file can be refactored with an object or more functions to depend less on this large set of globals.

igvc_navigation/src/path_follower/main.cpp
```diff
@@ -66,7 +66,10 @@ void position_callback(const nav_msgs::OdometryConstPtr& msg)
   RobotState cur_pos(msg);
 
   double goal_dist = cur_pos.distTo(waypoint->point.x, waypoint->point.y);
-  ROS_INFO_STREAM("Distance to waypoint: " << goal_dist << "(m.)");
+  if (debug)
+  {
+      ROS_INFO_STREAM("Distance to waypoint: " << goal_dist << "(m.)");
```
barulicm: If this information is only used for debug, consider using [ROS_DEBUG_STREAM](http://wiki.ros.org/roscpp/Overview/Logging). Using the built in log levels will allow you to turn debug information on and off without recompiling your code.

igvc_utils/include/igvc_utils/NodeUtils.hpp
```diff
@@ -29,11 +30,37 @@ void getParam(const ros::NodeHandle &pNh, const std::string &param_name, T &para
 /**
 Calculates euclidian distance between two points
 */
-inline double get_distance(double x1, double y1, double x2, double y2)
+template<typename T>
+inline T get_distance(T x1, T y1, T x2, T y2)
 {
   return sqrt(pow(x2 - x1, 2) + pow(y2 - y1, 2));
```
barulicm: Consider replacing with [std::hypot](https://en.cppreference.com/w/cpp/numeric/math/hypot)

igvc_utils/include/igvc_utils/NodeUtils.hpp
```diff
@@ -29,11 +30,37 @@ void getParam(const ros::NodeHandle &pNh, const std::string &param_name, T &para
...
+template<typename T>
+inline T get_distance(const std::tuple<T,T>& p1, const std::tuple<T,T>& p2)
+{
+    T x1,y1;
+    std::tie(x1,y1) = p1;
+
+    T x2,y2;
+    std::tie(x2, y2) = p2;
+
+    return sqrt(pow(x2 - x1, 2) + pow(y2 - y1, 2));
```
barulicm: If this is just wrapping the other `get_distance` with a different input type, consider just returning `get_distance(x1,y1,x2,y2)`

igvc_utils/include/igvc_utils/NodeUtils.hpp
```diff
@@ -29,11 +30,37 @@ void getParam(const ros::NodeHandle &pNh, const std::string &param_name, T &para
...
+}
+
+// symmetric round up
+// Bias: away from zero
+template <typename T>
+T ceil0( const T& value )
+{
+    T result = std::ceil( std::fabs( value ) );
+    return (value < 0.0) ? -result : result;
```
barulicm: Consider `return (value < 0.0) ? std::floor(value) : std::ceil(value);`

igvc_navigation/CMakeLists.txt
```diff
@@ -163,8 +163,6 @@ include_directories(
 #   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
 # )
 
-#############
-## Testing ##
```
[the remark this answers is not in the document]
Alescontrela: nope. re-inserted

igvc_navigation/include/igvc_navigation/Graph.h
```diff
@@ -0,0 +1,275 @@
...
+
+    // k_m, as defined in the D* lite paper, keeps track of the robot's movement
+    // in the grid space. Serves to increase each new node's key value by k_m as
+    // to maintain lower bounds in the priority queue
+    float K_M = 0;
+
+    /**
+    Default no-arg constructor
+    */
+    Graph();
```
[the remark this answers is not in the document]
Alescontrela: Oh cool, didn't know about this. I'll change this for Graph and DLitePlanner

igvc_navigation/include/igvc_navigation/Graph.h
```diff
@@ -0,0 +1,275 @@
...
+is 0.2 then each cell actually has side dimensions of 0.2m.
+
+Author: Alejandro Escontrela <aescontrela3@gatech.edu>
+Date Created: December 16, 2018
+*/
+
+#ifndef GRAPH_H
+#define GRAPH_H
+
+// #define NDEBUG # TODO: uncomment for production
```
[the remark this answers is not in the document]
Alescontrela: That explains the `NDEBUG re-defined` error I was getting when compiling with DC_MAKE_BUILD = release

igvc_navigation/include/igvc_navigation/Graph.h
```diff
@@ -0,0 +1,275 @@
...
+
+#include <igvc_msgs/map.h>
+
+#include <cv_bridge/cv_bridge.h>
+#include <utility>
+#include <vector>
+#include <tuple>
+#include <cmath>
+#include <limits>
+#include <assert.h>
```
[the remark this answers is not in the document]
Alescontrela: switched :+1:

igvc_navigation/include/igvc_navigation/Graph.h
```diff
@@ -0,0 +1,275 @@
...
+    */
+    cv_bridge::CvImagePtr Map;
+
+    Node Start; // start node in the search problem
+    Node Goal; // goal node in the search problem
+
+    // Updated cell information is used to update nodes that lie on the 4 corners of
+    // each updated cell. This is reset each time updateGraph is called. Each element
+    // is composed of <cell index, old value, new value>
+    std::vector<std::tuple<int,int>> updatedCells;
```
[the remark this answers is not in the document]
Alescontrela: Updated the documentation to better reflect functionality. I'm leaving it as a tuple just in case more dimensions need to be incorporated in the future. Also, I don't think a float tuple carries significantly more overhead than a pair.

igvc_navigation/src/path_follower/main.cpp
Alescontrela: I use the `debug` conditional to prevent the message from being output to stdout when output="screen" for the ros node. Seeing as ROS_DEBUG_STREAM still outputs to stdout, it might be worthwhile for me to leave it like this.

igvc_utils/include/igvc_utils/NodeUtils.hpp
Alescontrela: A classic case of reinventing the wheel. I'll implement this

igvc_utils/include/igvc_utils/NodeUtils.hpp
Alescontrela: I'll chain these methods

igvc_navigation/src/d_lite_planner/main.cpp
Alescontrela: I see these packages as being heavily dependent on one another, yet separate in their functionality and purpose. Trying to squash them together into a single object or a lighter-weight implementation might lead to a decrease in interpret-ability that isn't worth the marginally small performance increase i think this would bring.

igvc_navigation/src/path_follower/main.cpp
barulicm: The verbosity level of ROS can be set outside of the application such that DEBUG output does not print to the terminal.

Alescontrela: Oh that's really helpful. I'll switch over to this then.

igvc_navigation/src/d_lite_planner/main.cpp
barulicm: There is no real performance to be gained or lost here. The suggestion is based on concerns about safety and maintainability. These concerns stem from the difficulty that global variables introduce when it comes to tracking object lifetime and usage.

For example, `expanded_pub` is defined at global scope despite it only being used outside of `main` in `expanded_callback`. I, personally, would prefer that `expanded_pub` was simply declared in `main`, `expanded_callback` renamed to `publish_expanded_set`, and add the publisher as a parameter to the `publish_expanded_set` function. This makes the lifetime and usage of the publisher object more obvious, and the name change reflects the way the function is actually used. It is not a callback function, simply a utility function for converting and publishing the expanded nodes set. 

Similarly, `expanded_cloud` is defined globally despite being used exclusively inside of `expanded_callback`. (The only use in main is actually duplicate code setting the frame_id which `expanded_callback` already sets.) It would be clearer if the cloud was declared inside the function, filled with useful data, and then handed over to the publisher.

Furthermore, many of these variables are declared in global and only actually used in `main`! `CSpace`,`goal_range`,`path_pub`,`rateTime`,`follow_old_path`,`initialize_search`, and `publish_expanded` are all used exclusively in `main`. So moving those variables alone without addressing the other concerns would shrink the list of globals dramatically and make it more obvious to future readers which variables get used in multiple places.

igvc_navigation/include/igvc_navigation/Graph.h
```diff
@@ -0,0 +1,275 @@
...
+    space into account.
+
+    @param[in] cellInd index of cell whose val has changed
+    @return list of nodes who might be affected by changed cell value
+    */
+    std::vector<Node> getNodesAroundCellWithCSpace(const std::tuple<int,int>& cellInd);
+
+
+
+private:
```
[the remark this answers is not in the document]
Alescontrela: Ah, I thought I was going to use the private scope but didn't in the end. I'll remove

igvc_navigation/include/igvc_navigation/Graph.cpp
```diff
@@ -0,0 +1,490 @@
...
+            it_map++;
+        }
+    }
+}
+
+bool Graph::isValidNode(const Node& s)
+{
+    int x,y;
+    std::tie(x,y) = s.getIndex();
+    // with [row,col] indexing
```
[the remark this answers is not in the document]
Alescontrela: i'll stick to [row,col] since that makes more sense intuitively when working with CV::Mat

igvc_navigation/include/igvc_navigation/Graph.cpp
```diff
@@ -0,0 +1,490 @@
...
+}
+
+std::vector<Node> Graph::nbrs(const Node& s, bool include_invalid)
+{
+    std::vector<Node> neighbors;
+    int x,y;
+    std::tie(x,y) = s.getIndex();
+
+    // right
+    Node r(x+1, y);
```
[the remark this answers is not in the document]
Alescontrela: this is perfect. I'll switch over

igvc_navigation/src/d_lite_planner/CMakeLists.txt
Alescontrela: Because I use them for other planner implementations (i.e. Field D*). Do you think it would be better practice to have `include/src` and place the files in there?

igvc_navigation/include/igvc_navigation/PriorityQueue.h
```diff
@@ -0,0 +1,213 @@
+/**
+Priority Queue implementation for Field D* path planning algorithm.
```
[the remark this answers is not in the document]
Alescontrela: Yeah. It implements a priority queue to deal with the lexicographic order required by D* Lite and Field D*. It also implements a custom comparative to enforce  relative ordering. I'll put this in the header.

igvc_navigation/src/d_lite_planner/CMakeLists.txt
barulicm: "include" is generally for headers; "src" is generally for implementation files (*.cpp).

If you have common code between multiple executables, I'd make a library target in cmake. Give those files their own directory with appropriate "include/*.h" and "src/*.cpp" directories. Then individual executables can include the headers they need from that library and link against the library to get the definitions. 

If the common code is less substantial, you _can_ skip the library target, make the right folder structure, and simply add the common *.cpp files directly to each executable target, but this is arguably less beneficial to humanity.

igvc_navigation/include/igvc_navigation/PriorityQueue.h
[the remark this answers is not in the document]
barulicm: I may be missing something, but std::priority_queue lets you define your own comparator to order your priority queue however you please. Why does your custom sorting not fit into that mechanism?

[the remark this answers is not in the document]
oswinso: If I'm not mistaken the priority queue here is really just a sorted set with the methods renamed to match the priority queue ones, and not really a priority queue since it implements an arbitrary remove and contains

igvc_navigation/include/igvc_navigation/Graph.h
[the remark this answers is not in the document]
oswinso: I think a struct would be more helpful here to attach meaning to the different values, and it also can be easily modified

igvc_navigation/src/d_lite_planner/main.cpp
Alescontrela: Ahh ok I understand your point, and I agree. I'll limit the scope of all variables that shouldn't be in the global scope.

igvc_navigation/include/igvc_navigation/Graph.h
```diff
@@ -0,0 +1,275 @@
...
+    */
+    Graph();
+    ~Graph();
+
+    /**
+    Sets a value for the graph's configuration space
+
+    @param[in] CSpace a value for the configuration space
+    */
+    void setCSpace(float CSpace);
```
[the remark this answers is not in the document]
Alescontrela: Oh boy... this is embarrassing. I'll modify this PR to adhere to a single naming convention and make sure abbreviations don't get in the way of interpret-ability.

igvc_navigation/include/igvc_navigation/PriorityQueue.h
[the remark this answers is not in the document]
Alescontrela: I created this priority queue to implement the specific methods required by the incremental path planners. Some of these methods are those that would not be available in a regular `std::priority_queue`, i.e. `Node PriorityQueue::topNode()`, `PriorityQueue::topKey()`. Additionally, since the priority queue is specialized to handle nodes and their lexicographically-ordered keys, I defined the `Key` struct and corresponding Comparator to be used by the priority queue. In short, I find my specialization of `std::priority_queue` to make sense in the context of the graph search problem and as such, easier to interpret.

igvc_navigation/include/igvc_navigation/Graph.h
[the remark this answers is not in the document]
Alescontrela: I think a struct is overkill here. A vector of `std::tuple<float,float>` can be extended by adding more fields to the tuple. The vector only contain the indices of updated cells in the graph search problem, which can be inferred from the variable name `updatedCells`.

igvc_utils/include/igvc_utils/NodeUtils.hpp
Alescontrela: this is cleaner. I'll switch over.

igvc_navigation/include/igvc_navigation/Graph.h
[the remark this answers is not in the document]
oswinso: I don't think it is overkill, since a pair is essentially a struct, so if a struct was overkill, then by transitivity, a pair would also be overkill. The difference between the two is that, as @barulicm has said, the struct and it's members are named, which provides greater clarity to your code.

igvc_navigation/include/igvc_navigation/PriorityQueue.h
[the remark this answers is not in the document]
barulicm: After looking through the implementation of this yesterday, I agree with @oswinso. The things that make your container different (and justify its existence) are:
- the remove method
- the contains method

topKey and topNode are simply convenience methods over the standard library (which would look more like pq.top().key() or pq.top().node().

I want to stress that, while it is certainly useful for the graph search problem, the custom method of sorting elements is not something that justifies the creation of a new container. If this was the only thing different, you would be better off using standard components.

igvc_navigation/include/igvc_navigation/Graph.h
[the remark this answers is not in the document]
Alescontrela: `std::tuple` and `std::pair` are two different data structures; this code implements `std::tuple`, not `std::pair`. Anyways, let's say that I create a `Cell` struct. The only two fields in this struct would be `x` and `y`. This would hardly add any meaning given that the code is documented to reflect that each `std::tuple` in the vector is an <x,y> index for the cell.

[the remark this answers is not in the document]
barulicm: Semantic meaning expressed directly in the code is better than documentation comments.

tuple and pair aren’t all that different, and pair is largely regarded as a specialization of tuple. 

Structs would also help make it more obvious which parts of the code are handling the same data. There are a few spots in this code where there is a vector of tuples, structs would make it clearer if those vectors contained the same type or not.

Finally, “overkill” as measured by what? A struct is a very light and nimble tool which predates C and was created with the explicit purpose of grouping multiple data members together for convenient collective handling.

igvc_navigation/include/igvc_navigation/PriorityQueue.h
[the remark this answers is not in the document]
Alescontrela: I agree that the features my implementation adds over the standard priority_queue are light. But, these components are used extensively throughout the path planning process. Knowing this, I think using this container abstracts away functionality that would look out of place in, say, `DLitePlanner.cpp`. i.e.: 

```C++
PQ.remove(Node(2,2));
```

instead of:
```C++
std::set<std::pair<Node, Key>>::iterator it = std::find_if(
      pq.begin(), pq.end(), [n](std::pair<Node, Key> const& e) { return e.first.getIndex() == n.getIndex(); });

if (it == pq.end())  // no such item exists
   return false;

pq.erase(it);
```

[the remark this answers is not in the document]
barulicm: Yes, I am agreeing with you on this part. These are features that are not only more succinct in your implementation, but are straight up not possible in the standard container (std::priority_queue does not expose iterators, so there is no find or erase mechanism).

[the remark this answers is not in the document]
oswinso: Why are you using `std::find_if` and a lambda? 
Is this not simply
```c++
auto it = std::find(Node(2, 2));
if (it == pq.end()) {
  return false;
}
pq.erase(it);
```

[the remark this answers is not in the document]
barulicm: No, because he’s comparing pair objects using only the index member of the Node.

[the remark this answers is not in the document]
oswinso: oh, that's true. mb

igvc_navigation/include/igvc_navigation/Graph.h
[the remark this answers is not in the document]
Alescontrela: I agree with creating a `Cell` struct to differentiate between the data types contained within the various vectors in the code. I'll define it in `Graph.h`

igvc_navigation/include/igvc_navigation/Graph.h
```diff
@@ -0,0 +1,301 @@
...
+  the occupancy grid. Map is the current, most up-to-date occupancy grid.
+  */
+  cv_bridge::CvImagePtr Map;
+
+  Node Start;  // start node in the search problem
+  Node Goal;   // goal node in the search problem
+
+  // Updated cell information is used to update nodes that lie on the 4 corners of
+  // each updated cell. This is reset each time updateGraph is called. Each element
+  // is composed of an <x,y> tuple representing the cell.
```
oswinso: Make up your mind about you're using `/** */` or `//`, and on where you're placing the comment

igvc_navigation/include/igvc_navigation/Graph.h
```diff
@@ -0,0 +1,301 @@
...
+  */
+  float euclidianHeuristic(const std::tuple<int, int>& ind);
+  /**
+  Gets Nodes around a cell whose occupancy value has changed while taking configuration
+  space into account.
+
+  @param[in] cellInd index of cell whose val has changed
+  @return list of nodes who might be affected by changed cell value
+  */
+  std::vector<Node> getNodesAroundCellWithConfigurationSpace(const Cell& cell);
```
oswinso: @barulicm Is returning a std::vector directly fine here? I feel like it's better to return through a reference parameter.

igvc_navigation/include/igvc_navigation/Node.h
```diff
@@ -0,0 +1,148 @@
...
+};
+
+// ostream operator for Node
+std::ostream& operator<<(std::ostream& stream, const Node& n);
+
+// ostream operator for std::unordered_set of Node
+std::ostream& operator<<(std::ostream& stream, const std::unordered_set<Node>& uset);
+
+// Re-defined hash method for nodes. Used for maintaining unordered map.
+namespace std
```
oswinso: Why the namespace std? There's no reason to use it here imo

igvc_navigation/include/igvc_navigation/PriorityQueue.h
```diff
@@ -0,0 +1,216 @@
...
+  @param[in] n node to remove from the priority queue
+  @return whether such an item was found and removed
+  */
+  bool remove(Node n);  // O(n)
+  /**
+  Pops the top (least-cost) state from the priority queue. This has the effect of
+  reducing the size of the priority queue by one. If the priority queue is
+  empty, nothing happens.
+  */
+  void pop();  // O(1)
```
oswinso: This is O(log(n)), you're using a set, which uses a red-black tree as the backing data structure.

igvc_utils/include/igvc_utils/NodeUtils.hpp
```diff
@@ -29,9 +31,28 @@ void getParam(const ros::NodeHandle &pNh, const std::string &param_name, T &para
...
-  return sqrt(pow(x2 - x1, 2) + pow(y2 - y1, 2));
+  return std::hypot(x2 - x1, y2 - y1);
+}
+
+/**
+Calculates euclidian distance between two points, taking tuples for each
+(x,y) point as argument
+*/
+template<typename T>
+inline T get_distance(const std::tuple<T,T>& p1, const std::tuple<T,T>& p2)
```
oswinso: Proper documentation?

igvc_navigation/include/igvc_navigation/Node.h
Alescontrela: This is to specialize [std::hash](https://en.cppreference.com/w/cpp/utility/hash) for `Node`.

[Reference](https://stackoverflow.com/questions/8157937/how-to-specialize-stdhashkeyoperator-for-user-defined-type-in-unordered)

igvc_navigation/include/igvc_navigation/Graph.h
```diff
@@ -0,0 +1,301 @@
...
+In Field D*, Nodes are located on the corners of grid cells (as opposed to the
+center, like in A*, D*, or D*lite).
+
+The Graph object provides an interface for the occupancy grid, as well as nodes.
+To facilitate the search problem, the graph operates under a unit-cell assumption,
+meaning that each cell has side lengths of 1m. The `Resolution` parameter specifies
+the conversion between the unit cell and the actual cell dimensions. i.e. if Resolution
+is 0.2 then each cell actually has side dimensions of 0.2m.
+
+Author: Alejandro Escontrela <aescontrela3@gatech.edu>
```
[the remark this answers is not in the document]
Alescontrela: File comment blocks are pretty standard AFAIK

igvc_navigation/include/igvc_navigation/Graph.h
Alescontrela: This is something i'm always hesitant about. Any insight on this practice would be helpful.

igvc_navigation/include/igvc_navigation/PriorityQueue.h
Alescontrela: :+1: I'll update

Alescontrela:
```C++
auto it = this->find(n);

if (it == pq.end())  // no such item exists
  return false;

pq.erase(it);
return true;
```

and 

```C++
std::set<std::pair<Node, Key>>::iterator PriorityQueue::find(Node n)
{
  std::set<std::pair<Node, Key>>::iterator it = std::find_if(
      pq.begin(), pq.end(), [n](std::pair<Node, Key> const& e) { return e.first.getIndex() == n.getIndex(); });

  return it;
}
```

igvc_utils/include/igvc_utils/NodeUtils.hpp
Alescontrela: :+1: updated

igvc_navigation/include/igvc_navigation/Graph.h
Alescontrela: you're right. I'll update

igvc_navigation/include/igvc_navigation/Graph.h
Alescontrela: [This](https://stackoverflow.com/questions/15704565/efficient-way-to-return-a-stdvector-in-c) stack overflow post says that, in C++11, `std::vector<X> f();` is the preferred way to return a vector.

"With C++11, std::vector has move-semantics, which means the local vector declared in your function will be moved on return and in some cases even the move can be elided by the compiler."

igvc_navigation/src/d_lite_planner/CMakeLists.txt
Alescontrela: This definitely organized a code. I created a `STATIC` library under `igvc_navigation` and placed relevant files in there:
```
igvc_navigation
|...
|___ GraphSearch
    |___ include
    |   |___"Graph.h","Node.h","PriorityQueue.h"
    |___ src 
        |___"Graph.cpp","Node.cpp","PriorityQueue.cpp"

```
The `d_lite_planner` executable now links this library

igvc_navigation/include/igvc_navigation/Graph.h
barulicm: Even before C++11, you should never return a reference to a temporary variable. The vector returned here is created on demand, it's not something internal to the Graph object.

Even without move semantics, returning the plain vector type is the right thing to do.

Alescontrela:
Thanks for all the feedback!

PR #351, opened 2019-02-09
OcTree based Occpancy Grid Mapping

oswinso:
This PR replaces the old mapper with a new mapper which uses an OcTree backed occupancy grid mapper to map the surroundings based on data from the lidar, by mapping with known poses. The parameters are stored in octomap.yaml.

Currently, this mapper does not handle semantic segmentation data from the camera for line detection.

Todo:

- [ ] Handle semantic segmentation information about lines and potholes
- [ ] Incorporate the sensor model during update (ie. when objects are outside the detection range


JasonGibson274:
We should stop having todos on PR. If you still have work to do, don’t make a PR.

PR #353, opened 2019-02-12
added in apt and rosdep update steps

JasonGibson274:
Resolved issue where apt and rosdep not being updated when building container in circle ci.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ defaults: &image
   - image: robojackets/igvc-baseimage:latest
 
 defaults: &install_deps
-  run: rosdep install -iy --from-paths ./src
+  run: apt-get update && rosdep update && rosdep install -iy --from-paths ./src
 
 defaults: &save_src_cache
   save_cache:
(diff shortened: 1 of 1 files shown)


barulicm:
What's the current relationship between devel and master? Do you care that devel is 6 behind master including changes to this file?

JasonGibson274:
master is supposed to be code that is guaranteed to work together and should work well. devel is peer reviewed code that has been put together but may have  known issues or bugs, used to combine multiple dependent features updates that are not ready for master but should be tested together, but reviewed before that.

Devel is essentially acting as a master branch right now though.

Alescontrela:
Thoughts on merging devel into master after merging path planner and occupancy grid?

JasonGibson274:
Ugh now that I think about this we are kind of being dumb, we should just have temporary devel branches for specific things, like mapping and path planning.

JasonGibson274:
but yes

JasonGibson274:
@Alescontrela can I get a positive review?

PR #355, opened 2019-02-26
Field d

Alescontrela:
This PR contains the implementation for the Field D* incremental path planning algorithm. In contrast to D* Lite, which is restricted to headings of pi/4 radians, Field D* can travel to continuous points along edges. As such, the Field D* incremental path planning algorithm provides smooth paths in real time.

**Usage**:
```bash
$ roslaunch igvc_navigation field_d_planner.launch simulation:=true # terminal window 3
```
**References**:
[D* Lite Paper](http://idm-lab.org/bib/abstracts/papers/aaai02b.pdf)
[Field D* Paper](https://www.ri.cmu.edu/pub_files/pub4/ferguson_david_2005_3/ferguson_david_2005_3.pdf)


igvc_navigation/GraphSearch/include/Graph.h
```diff
@@ -27,10 +27,68 @@ Date Created: December 16, 2018
...
+  Position(std::tuple<float, float> position) : Position(std::get<0>(position), std::get<1>(position))
+  {
+  }
+
+  Position(Node n) : Position(static_cast<std::tuple<float, float>>(n.getIndex()))
+  {
+  }
+
+  // overloaded assignment operator
+  Position& operator=(const Position& other)
```
oswinso: Why do you explicitly define this, as well as `==` and `!=`? Pretty sure c++ by default implements these for you.

igvc_navigation/GraphSearch/include/Graph.h
```diff
@@ -100,6 +158,19 @@ class Graph
   // to maintain lower bounds in the priority queue
   float KeyModifier = 0;
 
+  float OccupancyThreshold = 0.7f;
```
oswinso: Magic Number.

igvc_navigation/GraphSearch/include/Graph.h
```diff
@@ -292,6 +366,9 @@ class Graph
   @return list of nodes who might be affected by changed cell value
   */
   std::vector<Node> getNodesAroundCellWithConfigurationSpace(const Cell& cell);
+
+private:
+  float OccupancyThresholdUChar = 178.5f;
```
oswinso: More magic numbers

igvc_navigation/GraphSearch/src/Graph.cpp
```diff
@@ -120,22 +124,76 @@ std::vector<Node> Graph::nbrs(const Node& s, bool include_invalid)
   int x, y;
   std::tie(x, y) = s.getIndex();
 
-  for (auto dx = -1; dx <= 1; dx++)
```
oswinso: Why remove this line and do the iterating manually?

igvc_navigation/GraphSearch/include/Graph.h
```diff
@@ -27,10 +27,68 @@ Date Created: December 16, 2018
...
+Position differs from a Node in that it is not restricted to lying on a vertex, but
+can also lie along an edge. As such, a position is represented by two floats <x,y>
+that represent the position's x & y coordinates in the Graph.
+*/
+struct Position
+{
+  float x;
+  float y;
+
+  Node castToNode() const
```
oswinso: camelCase here

igvc_navigation/GraphSearch/src/Graph.cpp
```diff
@@ -385,25 +445,20 @@ float Graph::getTraversalCost(const Node& s, const Node& s_prime)
...
-  float x_prime, y_prime;
-  std::tie(x_prime, y_prime) = p_prime;
-
   // round p to get a reference node
-  Node s(std::make_tuple(static_cast<int>(roundf(x)), static_cast<int>(roundf(y))));
+  Node s(std::make_tuple(static_cast<int>(roundf(p.x)), static_cast<int>(roundf(p.y))));
 
   // get relative distance between p_prime and p. Note: ceil0 is biased away from 0
   // i.e. ceil0(-0.35) = -1.00
-  float x_diff = igvc::ceil0(x_prime - x);
```
oswinso: snake_case here

igvc_navigation/GraphSearch/src/Graph.cpp
```diff
@@ -431,19 +486,39 @@ float Graph::euclidianHeuristic(const std::tuple<int, int>& ind)
...
-  int sep = static_cast<int>(ConfigurationSpace / Resolution);  // number of cells on all sides that constitute C-space
+  std::queue<Node> openList;           // nodes to evaluate
+  std::unordered_set<Node> closedSet;  // evaluated nodes
+  std::vector<Node> cellNodes;         // nodes that require update
 
-  std::vector<Node> cellNodes;
+  Node startNode(cell.x, cell.y);
 
-  for (int new_x = cell.x - sep; new_x <= cell.x + sep + 1; new_x++)
+  openList.push(startNode);
```
oswinso: camelCase here

igvc_navigation/src/d_lite_planner/DLitePlanner.cpp
```diff
@@ -73,7 +73,8 @@ int DLitePlanner::computeShortestPath()
 
   int numNodesExpanded = 0;
 
-  while ((PQ.topKey() < calculateKey(NodeGrid.Start)) || (getRHS(NodeGrid.Start) != getG(NodeGrid.Start)))
+  while (((PQ.topKey() < calculateKey(NodeGrid.Start)) || (getRHS(NodeGrid.Start) != getG(NodeGrid.Start))) &&
+         (PQ.size() > 0))
```
oswinso: change to `!PQ.empty()`, more clear

igvc_navigation/src/field_d_planner/FieldDPlanner.h
```diff
@@ -0,0 +1,254 @@
...
+
+  CostComputation(float x, float y, float cost)
+  {
+    this->x = x;
+    this->y = y;
+    this->cost = cost;
+  }
+
+  // overloaded assignment operator
+  CostComputation& operator=(const CostComputation& other)
```
oswinso: unnecessary overload,

igvc_navigation/src/field_d_planner/FieldDPlanner.h
```diff
@@ -0,0 +1,254 @@
...
+{
+public:
+  // Graph contains methods to deal with Node(s) as well as updated occupancy
+  // grid cells
+  Graph NodeGrid;
+
+  std::vector<Position> Path;
+
+  // path additions made by one step of constructOptimalPath()
+  typedef std::pair<std::vector<Position>, float> path_additions;
```
oswinso: prefer `using` over typedef

igvc_navigation/src/field_d_planner/FieldDPlanner.h
```diff
@@ -0,0 +1,254 @@
...
+  bool isWithinRangeOfGoal(const Position& p);
+  /**
+  Tries to insert an entry into the unordered map. If an entry for that key
+  (Node) already exists, overrides the value with specified g and rhs vals.
+
+  @param[in] s key to create entry for
+  @param[in] g g value for entry
+  @param[in] rhs rhs value for entry
+  */
+  void insert_or_assign(Node s, float g, float rhs);
```
oswinso: again, make up your mind snake_case or camelCase. I think you should convert everything to snake_case.

igvc_navigation/src/field_d_planner/FieldDPlanner.h
```diff
@@ -0,0 +1,254 @@
...
+  Returns list of indices of expanded nodes, that is, the indices of the
+  nodes contained in the unordered map.
+
+  @return vector of nodes in the unordered map
+  */
+  std::vector<std::tuple<int, int>> getExplored();
+
+private:
+  // hashed map contains all nodes and <g,rhs> values in search
+  std::unordered_map<Node, std::tuple<float, float>> umap;
```
oswinso: better variable names?

igvc_navigation/src/field_d_planner/FieldDPlanner.h
```diff
@@ -0,0 +1,254 @@
...
+  // priority queue contains all locally inconsistent nodes whose values
+  // need updating
+  PriorityQueue PQ;
```
oswinso: why the all caps ABBRV?

igvc_navigation/src/field_d_planner/main.cpp
```diff
@@ -0,0 +1,268 @@
...
+//--------------------------- ROS Callbacks ----------------------------//
+
+/**
+    Set the current map to be used by the D* Lite search problem. The initial
+    map is used to perform the first search through the occupancy grid (equivalent
+    to A*). All maps thereafter are used to update edge costs for the search problem.
+*/
+void map_callback(const igvc_msgs::mapConstPtr& msg)
+{
+  std::lock_guard<std::mutex> planning_lock(planning_mutex);
```
oswinso: not necessary since this is single threaded.

igvc_navigation/src/field_d_planner/main.cpp
```diff
@@ -0,0 +1,268 @@
...
+*/
+void waypoint_callback(const geometry_msgs::PointStampedConstPtr& msg)
+{
+  std::lock_guard<std::mutex> lock(planning_mutex);
+
+  int goal_x, goal_y;
+  goal_x = static_cast<int>(std::round(msg->point.x / planner.NodeGrid.Resolution)) + x_initial;
+  goal_y = static_cast<int>(std::round(msg->point.y / planner.NodeGrid.Resolution)) + y_initial;
+
+  std::tuple<int, int> new_goal = std::make_tuple(goal_x, goal_y);
```
oswinso: didn't you make a struct `Position`?

igvc_navigation/src/field_d_planner/main.cpp
```diff
@@ -0,0 +1,268 @@
...
+  planner.NodeGrid.setGoal(new_goal);
+
+  float distance_to_goal = planner.NodeGrid.euclidianHeuristic(new_goal) * planner.NodeGrid.Resolution;
+
+  ROS_INFO_STREAM((goal_changed ? "New" : "Same") << " waypoint received. Search Problem Goal = " << goal_x << ", "
+                                                  << goal_y << ". Distance: " << distance_to_goal << "m.");
+
+  if (distance_to_goal > maximum_distance)
+  {
+    ROS_WARN_STREAM("Planning to waypoint more than " << maximum_distance
```
oswinso: Change to `ROS_WARN_STREAM_THROTTLE` so the terminal isn't spammed

igvc_navigation/src/field_d_planner/main.cpp
```diff
@@ -0,0 +1,268 @@
...
+      ROS_INFO_STREAM(num_nodes_updated << " nodes updated");
+
+    // only update the graph if nodes have been updated
+    if ((num_nodes_updated > 0) || initialize_search)
+    {
+      ros::Time begin = ros::Time::now();
+      num_nodes_expanded = planner.computeShortestPath();
+
+      double elapsed = (ros::Time::now() - begin).toSec();
+      ROS_INFO_STREAM(num_nodes_expanded << " nodes expanded in " << elapsed << "s.");
```
oswinso: DEBUG?

igvc_navigation/src/field_d_planner/main.cpp
```diff
@@ -0,0 +1,268 @@
...
+    planner.constructOptimalPath(lookahead_dist);
+
+    if (publish_expanded)
+      publish_expanded_set(planner.getExplored(), expanded_cloud, expanded_pub);
+
+    nav_msgs::Path path_msg;
+    path_msg.header.stamp = ros::Time::now();
+    path_msg.header.frame_id = "odom";
+
+    for (Position pos : planner.Path)
```
oswinso: Move this to a separate method? this while loop is very long.

igvc_rviz_plugins/configs/jessii_conf.rviz
```diff
@@ -12,7 +12,7 @@ Panels:
         - /PointCloud21
         - /PointCloud22
       Splitter Ratio: 0.5
-    Tree Height: 667
+    Tree Height: 258
```
oswinso: Probably don't commit this file.

igvc_navigation/src/field_d_planner/main.cpp
```diff
@@ -0,0 +1,268 @@
...
+    {
+      geometry_msgs::PoseStamped pose;
+      pose.header.stamp = path_msg.header.stamp;
+      pose.header.frame_id = path_msg.header.frame_id;
+      pose.pose.position.x = (pos.x - x_initial) * planner.NodeGrid.Resolution;
+      pose.pose.position.y = (pos.y - y_initial) * planner.NodeGrid.Resolution;
+      path_msg.poses.push_back(pose);
```
oswinso: I think it would be better to also include the heading of the pose here

igvc_navigation/GraphSearch/include/Graph.h
Alescontrela: I don't think this is correct. How would C++ know what fields I want to compare for equality? The default `==` operator checks for reference equality if i'm not mistaken.

igvc_navigation/GraphSearch/include/Graph.h
Alescontrela: This is the default value but this value can be overridden with a launch parameter

igvc_navigation/GraphSearch/include/Graph.h
Alescontrela: Same as above. This value is just `occupancy_threshold *255.0f`

igvc_navigation/GraphSearch/src/Graph.cpp
Alescontrela: This is because the order in which neighboring nodes are returned is important for Field D*. They are returned in a counterclockwise fashion starting at the right node

igvc_navigation/src/field_d_planner/main.cpp
Alescontrela: Done :+1:

igvc_navigation/GraphSearch/include/Graph.h
oswinso: Oh woops, for some reason I was under the impression that C++ had default struct comparison for some reason.

igvc_navigation/src/field_d_planner/main.cpp
Alescontrela: yep :+1:

igvc_navigation/GraphSearch/include/Graph.h
oswinso: I think it'd be better to have this initialized using igvc::param? So that if someone deleted the default param by accident, we'd get a warning from launching the Node

igvc_navigation/GraphSearch/src/Graph.cpp
oswinso: perhaps refactor this out into another method for clarity?

Alescontrela: It would be a bit roundabout to create another method that effectively performs this method's functionality. i.e. 

```c++
std::vector<Node> Graph::nbrs(const Node& s, bool include_invalid) {
    return nbrs2(s, include_invalid);
}
```

oswinso: I mean like, right below this method is basically another method that does the same thing, which I feel like could be combined.

igvc_navigation/src/field_d_planner/main.cpp
```diff
@@ -0,0 +1,266 @@
...
+  expanded_cloud.clear();
+  expanded_cloud.header.frame_id = "odom";
+
+  for (std::tuple<int, int> ind : inds)
+  {
+    pcl::PointXYZRGB p;
+
+    p.x = static_cast<float>(std::get<0>(ind) - x_initial) * map->resolution;
+    p.y = static_cast<float>(std::get<1>(ind) - y_initial) * map->resolution;
+    p.z = -0.05f;
```
jzheng84: Is this arbitrary?

igvc_navigation/GraphSearch/src/Graph.cpp
Alescontrela: The method below `nbrs`, `nbrsContinuous`, returns a different data structure altogether. I've considered ways of reducing the size of the code by grouping various methods together but I haven't figured out a good way of doing so without complicating the codebase significantly.

igvc_navigation/src/field_d_planner/main.cpp
Alescontrela: Yeah the `p.z = -0.05f` is arbitrary. I didn't want the visualization of expanded nodes to conflict with the path visualization so I just put the expanded pointcloud below.

igvc_navigation/src/d_lite_planner/DLitePlanner.cpp
Alescontrela: :+1:

igvc_navigation/src/field_d_planner/FieldDPlanner.h
Alescontrela: changed to `ExpandedMap`

igvc_navigation/src/field_d_planner/FieldDPlanner.h
Alescontrela: Changing this to `PriorityQueue`

igvc_navigation/src/field_d_planner/main.cpp
Alescontrela: This would significantly change the search problem and is outside the scope of this PR

igvc_navigation/GraphSearch/include/Graph.h
Alescontrela: i'm currently initializing it with `igvc::getParam`  so an exception is thrown if the parameter isn't supplied.

igvc_navigation/src/field_d_planner/main.cpp
Alescontrela: A better data structure to use in this case is `Node` since the goal will always be a node lying along the graph's vertices. Implemented.

igvc_navigation/src/field_d_planner/main.cpp
Alescontrela: I might create a wrapper class which takes in a `NodeHandle` and runs the path planner. Thoughts on this? Similar to [this](https://github.com/RoboJackets/igvc-software/blob/feat/refactor_smooth_control/igvc_navigation/src/path_follower/Path_follower.cpp)

igvc_navigation/src/field_d_planner/main.cpp
oswinso: I mean like including the heading from the current node to the next node with a simple atan2 of the difference in position, so that smooth control does not need to calculate it again

igvc_navigation/src/field_d_planner/FieldDPlanner.h
Alescontrela: I named this `insert_or_assign` because as of c++17, `std::unordered_map` has a member function called [`insert_or_assign`](https://en.cppreference.com/w/cpp/container/unordered_map/insert_or_assign). However, we still compile with c++11 and cannot make use of this feature so I created a method that implements this functionality with the same name. It does indeed look a bit out of place.

igvc_navigation/GraphSearch/src/Graph.cpp
Alescontrela: Addressed in latest commit

igvc_navigation/GraphSearch/include/Graph.h
Alescontrela: Addressed in latest commit

igvc_navigation/src/field_d_planner/FieldDPlanner.h
Alescontrela: Naming convention issues addressed in latest commit

igvc_navigation/src/field_d_planner/main.cpp
oswinso: I guess this can be an issue that is resolved later.

igvc_navigation/src/field_d_planner/main.cpp
oswinso: Yeah, I think that'd be a nice refactor, to separate the ROS stuff from the actual algorithm.

igvc_navigation/src/field_d_planner/main.cpp
Alescontrela: implemented :+1:

igvc_navigation/src/field_d_planner/FieldDPlanner.cpp
```diff
@@ -0,0 +1,637 @@
...
+
+  float g_p1 = getEdgePositionCost(p1);  // path cost of nearest neighbor
+  float g_p2 = getEdgePositionCost(p2);  // path cost of diagonal neighbor
+
+  float d_p1 = 1.0f;  // distance to nearest neighbor
+  assert(d_p1 > 0.0f);
+  float d_p2 = sqrtf(2);  // distance to diagonal
+  assert(d_p2 > 0.0f);
+  float d_n = 1.0f;  // distance between consecutive neighbors (edge length)
+  assert(d_n > 0.0f);
```
oswinso: I don't think these three asserts do much

Alescontrela: Yeah they don't

igvc_navigation/src/field_d_planner/FieldDPlanner.cpp
```diff
@@ -136,10 +135,17 @@ void FieldDPlanner::publish_expanded_set(pcl::PointCloud<pcl::PointXYZRGB>& expa
       p.g = 125;
       p.b = static_cast<uint8_t>((std::get<0>(e.second) / max_g) * 255.0f);
     }
-    expanded_cloud.points.push_back(p);
+    expanded_cloud->points.push_back(p);
   }
 
-  expanded_pub_.publish(expanded_cloud);
+  // pass cloud through a VoxelGrid to reduce size
+  pcl::VoxelGrid<pcl::PointXYZRGB> sor;
```
oswinso: I'm not sure about whether we need to reduce the size of this. If you look at the rosbag for one particular bag file, `expanded` only has like 571 messages, while all 6 variants of the map pointclouds have like 2000 messages.

Alescontrela: I agree. I implemented this mostly for visualization tuning but it doesn't add much. I'm going to revert to 5228adb, remove some unnecessary asserts, then merge

PR #363, opened 2019-03-17
Added throttling to logging

oswinso:
This PR adds throttling to logging messages in `motor_controller` by changing `ROS_INFO_STREAM` to `ROS_INFO_STREAM_THROTTLE`, and adds a parameter `log_period` to the launch file.


oswinso:
Changed rate to an actual rate by doing `1/dt`.

PR #364, opened 2019-03-17
Ethernet 

Alescontrela:
These changes implement the code required to allow the main computer to communicate with the mbed microcontroller via ethernet.


igvc_utils/include/igvc_utils/EthernetSocket.h
```diff
@@ -0,0 +1,82 @@
...
+  /**
+  Get version of boost used by this library
+
+  @return version number in major_version.minor_version.patch_level format
+  */
+  std::string getBoostVersion();
+
+private:
+  boost::asio::io_service io_service;                  // provides core io functionality
+  std::unique_ptr<boost::asio::ip::tcp::socket> sock;  // tcp connection socket
```
oswinso: `io_service_` and `sock_` here.

igvc_platform/src/motor_controller/main.cpp
```diff
@@ -1,247 +1,286 @@
...
-      else if (pound != std::string::npos)
+
+      /* Send the message strapped to a pigeon's leg! */
+      sock.sendMessage(reinterpret_cast<char*>(requestbuffer), message_length);
+
+      /* Read response from the server */
+      size_t n; // n is the response from socket: 0 means connection closed, otherwise n = num bytes read
+      uint8_t buffer[256]; // buffer to read response into
+
+      memset(buffer, 0, sizeof(buffer));
```
oswinso: I think `uint8_t buffer[256] = {0};` is cleaner than an explicit memset.

igvc_platform/src/motor_controller/main.cpp
```diff
@@ -1,247 +1,286 @@
...
+      /* read from the buffer */
+      n = sock.readMessage(buffer); // blocks until data is read
+
+      if (n == 0)
```
oswinso: This method is really long. Refactor it into smaller methods.

igvc_platform/src/motor_controller/main.cpp
```diff
@@ -1,247 +1,286 @@
...
+    // memcpy(responsebuffer, buffer, sizeof(responsebuffer));
+
+    if (n == 0)
+    {
+        ROS_ERROR_STREAM("Connection closed by server");
+        ros::shutdown();
+    }
+
+    /* Allocate space for the decoded message. */
+    ResponseMessage response = ResponseMessage_init_zero;
```
oswinso: Doesn't the default constructor take care of this?

Alescontrela: I don't think a default constructor is defined for `ResponseMessage`. Also, I think it's better to make it explicit that all fields will be initialized to zero.

igvc_platform/src/motor_controller/main.cpp
Alescontrela: I think i'll leave it as `memset` as I find it easier to understand in this case

igvc_utils/include/igvc_utils/EthernetSocket.h
Alescontrela: :+1: done

igvc_platform/src/motor_controller/main.cpp
Alescontrela: The way this loop is structured currently is as follows: 

1. create a `RequestMessage` and send it to the microcontroller using the "/motors" topic
2. receive a `ResponseMessage` and publish its contents to "/battery", "/robot_enabled", and "/encoders"

I don't think this loop would benefit greatly from being refactored. I say this primarily because there would be a trade-off between decreasing the loop complexity and introducing more global variables. For example, if I were to create a `recieveResponse(...)` method, I would have to either a) make three of the above-mentioned publishers global variables or b) pass the publishers by reference which is bad for scalability.

Alescontrela: On second thought, i'm going to refactor this entire implementation into a class

Alescontrela: addressed

Alescontrela:
Will first finalize tests on robot before merging

igvc_platform/src/motor_controller/motor_controller.cpp
```diff
@@ -168,101 +164,102 @@ void MotorController::setPID()
...
-      ROS_ERROR_STREAM_THROTTLE(log_period_, "Battery voltage dangerously low:"
-                                                << "\n\tCurr. Voltage: " << battery_avg_
-                                                << "\n\tMin. Voltage: " << min_battery_voltage_);
-    }
+  if (battery_avg_ < min_battery_voltage_)
+  {
+    ROS_ERROR_STREAM_THROTTLE(log_period_, "Battery voltage dangerously low:"
+                                               << "\n\tCurr. Voltage: " << battery_avg_
+                                               << "\n\tMin. Voltage: " << min_battery_voltage_);
+  }
```
oswinso: This could probably also be refactored into a separate method for visual clarity

Alescontrela: addressed

Alescontrela:
Still works on robot with new changes.

Issue #369, opened 2019-03-21
Refactor mapper


oswinso:
The current mapper is horribly structured, with both `Mapper` and `Octomapper` taking on multiple responsibilities, with very long methods. These two classes should be refactored to be much smaller.

oswinso:
Closing for now, as the organization is slightly better, but I've opened a new issue #391 that more specifically deals with more clearly defining the identity of `octomapper`.

PR #370, opened 2019-03-21
Probabilistic Mapper

oswinso:
This PR uses [octomap ros](http://wiki.ros.org/octomap) to create an occupancy grid mapping solution that probabilistically maps both lidar scans and segmented lines from camera images.

The code in this branch is a bit messy, to be resolved with #369. However, since the old mapper hasn't been used for a long time, I think that we should do a PR first in terms of functionality, and then do a later PR for refactoring this class to be much cleaner.


oswinso:
Closing as I have since refactored this.

PR #372, opened 2019-03-21
Parameter assertions

oswinso:
This PR adds an assertion feature to `igvc::param` and `igvc::getParam` to add primitive assertions to the parameter values such as `Assertion::POSITIVE` and `Assertion::NEGATIVE`, and also changes the warning when a default value is used for `igvc::param` from `ROS_ERROR` to `ROS_WARN`.


Alescontrela:
Could you briefly describe some use cases for including `Assertion::POSITIVE` and `Assertion::NEGATIVE` in the function call?

oswinso:
I guess one use case would be if somebody accidentally adds a negative sign to some parameter which shouldn't be negative (ie. axle length), which would then mess up some calculation somewhere that assumes that this value is positive. For example, motor velocities are calculated using the axle length. If the axle length was accidentally set to a negative value, then all velocities would also become negative.

igvc_utils/include/igvc_utils/NodeUtils.hpp
```diff
@@ -1,31 +1,131 @@
...
+
+template <class T>
+void param(const ros::NodeHandle &pNh, const std::string &param_name, T &param_val, T &&default_val,
+           Assertion assertion = Assertion::NONE)
 {
   if (!pNh.param(param_name, param_val, default_val))
   {
-    ROS_ERROR_STREAM("Missing parameter " << param_name << " from " << pNh.getNamespace()
-                                          << ". Continuing with default values " << default_val);
+    ROS_WARN_STREAM("[" << pNh.getNamespace() << "] Missing parameter " << param_name
```
Alescontrela: Compiler error thrown here when `default_val` doesn't have an overloaded `<<` operator

Alescontrela:
I don't understand some of the logic in this PR:

Following the flow of a call to `igvc::getParam(const ros::NodeHandle &pNh, const std::string &param_name, T &param_val, Assertion assertion = Assertion::NONE)` where `assertion = Assertion::POSITIVE` and the param is indeed specified and positive (be it a float, double, int, etc...):

```C++
if (!pNh.getParam(param_name, param_val))
{
    ROS_ERROR_STREAM("[" << pNh.getNamespace() << "] Missing parameter " << param_name << ". Exiting...");
    ros::shutdown();
}
else
{
    check_assertion(assertion, pNh.getNamespace(), param_val, param_name);
}
```

Since the param is retrieved, `pNh.getParam(...)` returns `true` and the conditional evaluates to `false`, so `check_assertion(assertion, pNh.getNamespace(), param_val, param_name);` gets called. 

In `check_assertion(...)`:

```C++
switch (assertion)
  {
    case Assertion::POSITIVE:
      assert_positive(node_namespace, variable, variable_name);
      break;
    case Assertion::NEGATIVE:
      assert_negative(node_namespace, variable, variable_name);
      break;
    case Assertion::NONE:
      break;
  }
```

`Assertion::POSITIVE` was passed in therefore `assert_positive(node_namespace, variable, variable_name);` gets called

in `assert_positive`:
```C++
template <class T>
inline void assert_positive(const std::string &node_namespace, T &variable, const std::string &variable_name)
{
  ROS_ERROR_STREAM("[" << node_namespace << "] " << variable_name << " (currently " << variable
                       << ") should be greater than 0. Exiting...");
  ros::shutdown();
}
```

I hope you see my point. Even though the parameter is positive, no check is made to determine whether it is indeed positive or negative and `ros::shutdown();` is called regardless.

igvc_utils/include/igvc_utils/NodeUtils.hpp
oswinso: Ddin't fix this, but I added a template specialization for `std::vector`, which should be good enough for now.

oswinso:
Fixed the bugs you mentioned.

Alescontrela:
These changes don't compile. I think you're checking that a `std::vector` is greater and less an an, which is not an overloaded operation:

```bash
/home/aescontrela3/catkin_ws/src/igvc-software/igvc_utils/include/igvc_utils/NodeUtils.hpp:91:58: note:   mismatched types ‘const std::vector<_Tp, _Alloc>’ and ‘int’
   assert_with_default(node_namespace, variable, variable < 0, std::forward<T>(default_value), message.str());
```

oswinso:
Yeah, weird that merging in devel broke it lol. I fixed the changes + added the ability to define your own assertions via a lambda function, ie.
```c++
  igvc::getParam(pNh, "occupancy_threshold", occupancy_threshold_, [](float x){ return x > 0.0 && x < 1.0;});
```
Screenshot from 2019-03-31 15-31-25

oswinso:
@Alescontrela Fixed it.

Alescontrela:
Any idea why tests are failing?

oswinso:
Nope. Works on my machine.

Issue #365, opened 2019-03-21
Smooth control should path to a single target position


oswinso:
Currently, smooth control tries to path to a position `lookahead_distance_` away from the current position, which probably is not Lyapunov stable, and is different from the use case detailed in the paper. Ideally, by changing our target to not move until either we reach it, or the target point changes due to the path drastically changing, our trajectories should converge towards the path.

PR #373, opened 2019-03-22
Path follower pathing to "mini-waypoints" within the path for Lyapunov stability

oswinso:
This PR changes the path follower to path to "mini-waypoints" within the path instead of pathing towards a constantly moving target (which may not guarantee Lyapunov stability), addressing #365, as well as smoothing the path generated by taking the midpoints only, so that sharp turns are made to be more gradual.

example_old
Example scenario where the old path planner doesn't really converge on the path.
<br/>
example_new
Same scenario, but with the new planner, which does converge onto the path.

### DIFF
--- a/igvc_navigation/launch/follower.launch
+++ b/igvc_navigation/launch/follower.launch
@@ -1,0 +1,22 @@
+<?xml version="1.0"?>
+<launch>
+    <!-- Follower -->
+    <node name="path_follower" pkg="igvc_navigation" type="path_follower" output="screen" >
+        <!-- Maxmimum wheel velocity allowed before stopping motors -->
+        <param name="maximum_vel" type="double" value="2"/>
+        <param name="axle_length" type="double" value="0.52"/>
+        <!-- Distance along the path to generate smooth control law for  -->
+        <param name="lookahead_dist" type="double" value="1.3"/>
+        <!-- Frequency (hz) at which simulation is done -->
+        <param name="simulation_frequency" type="double" value="10.0"/>
+        <!-- How many seconds in the future should be simulated-->
+        <param name="simulation_horizon" type="double" value="5.0"/>
+        <!-- How far away robot should be from the waypoint before it stops -->
+        <param name="stop_dist" type="double" value="0.7"/>
+        <!-- Smooth Control Parameters -->
+        <param name="target_v" type="double" value="0.4"/>
+        <param name="k1" type="double" value="1.8"/>
+        <param name="k2" type="double" value="5"/>
+        <param name="debug" type="bool" value="true"/>
+    </node>
+</launch>
--- a/igvc_navigation/launch/follower.launch
+++ b/igvc_navigation/launch/follower.launch
@@ -18,7 +18,7 @@
         <!-- Smooth Control Parameters -->
         <param name="target_v" type="double" value="0.4"/>
         <param name="k1" type="double" value="1.8"/>
-        <param name="k2" type="double" value="2"/>
+        <param name="k2" type="double" value="5"/>
         <param name="debug" type="bool" value="true"/>
     </node>
 </launch>
--- a/igvc_navigation/launch/follower.launch
+++ b/igvc_navigation/launch/follower.launch
@@ -6,17 +6,19 @@
         <param name="maximum_vel" type="double" value="2"/>
         <param name="axle_length" type="double" value="0.52"/>
         <!-- Distance along the path to generate smooth control law for  -->
-        <param name="lookahead_dist" type="double" value="1.3"/>
+        <param name="lookahead_dist" type="double" value="0.4"/>
         <!-- Frequency (hz) at which simulation is done -->
         <param name="simulation_frequency" type="double" value="10.0"/>
         <!-- How many seconds in the future should be simulated-->
         <param name="simulation_horizon" type="double" value="5.0"/>
+        <param name="target_reached_distance" type="double" value="0.1"/>
+        <param name="target_move_threshold" type="double" value="0.1"/>
         <!-- How far away robot should be from the waypoint before it stops -->
         <param name="stop_dist" type="double" value="0.7"/>
         <!-- Smooth Control Parameters -->
         <param name="target_v" type="double" value="0.4"/>
         <param name="k1" type="double" value="1.8"/>
-        <param name="k2" type="double" value="5"/>
+        <param name="k2" type="double" value="2"/>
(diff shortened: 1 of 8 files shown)


Alescontrela:
After analyzing these changes in sim for some iterations it's clear that this path follower update decreases the "gracefulness" of the Smooth Control Law as defined in the original paper: 

> Thus we require the motion of the robot to be graceful.
> We use the qualitative term graceful as introduced by [1] to
> refer to a motion visibly safe, comfortable, fast and intuitive.
> To be safe and comfortable, a motion needs to be smooth
> with bounded velocity, acceleration and jerk.

I make this conclusion because the angular velocities produced by the Smooth Control law are much larger and, as such, jerkier. This is evident in the runtime errors that cause the robot to stop: 

> [ERROR] [1553551818.295902998, 344.341000000]: Maximum velocity exceeded. Right: -2.34842(m/s), Left: 3.14842(m/s), Max: 3(m/s) ... Stopping robot...

and, at times, velocities are generated that don't make sense: 

> [ERROR] [1553551824.839819656, 349.350000000]: Maximum velocity exceeded. Right: -17.9007(m/s), Left: 18.7007(m/s), Max: 3(m/s) ... Stopping robot...

I'll still review your changes to see if I can spot any obvious errors but, as of now, I don't think this PR improves the current path follower implementation.

igvc_navigation/launch/follower.launch
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
```
Alescontrela: I should add that, even with these launch parameters, I still observe the jerky behavior mentioned in my[ first comment ](https://github.com/RoboJackets/igvc-software/pull/373#issuecomment-476398552) for this PR.

igvc_navigation/launch/follower.launch
```diff
@@ -0,0 +1,24 @@
...
+        <!-- Maxmimum wheel velocity allowed before stopping motors -->
+        <param name="maximum_vel" type="double" value="2"/>
+        <param name="axle_length" type="double" value="0.52"/>
+        <!-- Distance along the path to generate smooth control law for  -->
+        <param name="lookahead_dist" type="double" value="0.4"/>
+        <!-- Frequency (hz) at which simulation is done -->
+        <param name="simulation_frequency" type="double" value="10.0"/>
+        <!-- How many seconds in the future should be simulated-->
+        <param name="simulation_horizon" type="double" value="5.0"/>
+        <param name="target_reached_distance" type="double" value="0.1"/>
```
Alescontrela: Also, add short descriptions that describe these parameters (`target_reached_distance` and `target_move_threshold`)

igvc_navigation/src/path_follower/smooth_control.cpp
```diff
@@ -126,81 +127,129 @@ Action SmoothControl::getAction(const RobotState& state, const RobotState& targe
...
+  // If already have a target, find the closest point along the current path to the current target
+  // If the closest point is too far,then break and find a new target
+  if (target)
+  {
+    // If we haven't reached the target
+    if (!reachedTarget(state, *target))
+    {
+      size_t target_index = findTargetOnPath(path, *target);
+      RobotState newTarget{ path->poses[target_index] };
+      // If the target is found, and the target isn't behind us, and the target is closer to target + 1 than we are
```
Alescontrela: comment ends too soon

igvc_navigation/src/path_follower/path_follower.cpp
```diff
@@ -157,17 +164,43 @@ void PathFollower::positionCallback(const nav_msgs::OdometryConstPtr& msg)
...
-  size_t num_poses = new_path->poses.size();
-  for (int i = 0; i < num_poses; i++)
+  // Patch heading into each point
+  nav_msgs::Path new_path = *msg;
+  size_t num_poses = new_path.poses.size();
+  for (size_t i = 0; i < num_poses; i++)
   {
-    double delta_x = new_path->poses[i + 1].pose.position.x - new_path->poses[i].pose.position.x;
-    double delta_y = new_path->poses[i + 1].pose.position.y - new_path->poses[i].pose.position.y;
+    double delta_x = new_path.poses[i + 1].pose.position.x - new_path.poses[i].pose.position.x;
```
Alescontrela: I think that calculating the headings for the original path, then calculating a new path, `smoothed_path`, then calculating the new headings for that path is a bit wasteful.

Why not make `smoothed_path` your only path to begin with and avoid calculating headings for the original path.

igvc_navigation/src/path_follower/smooth_control.cpp
```diff
@@ -126,81 +127,129 @@ Action SmoothControl::getAction(const RobotState& state, const RobotState& targe
...
+  , target_reached_distance_{ target_reached_distance }
+  , target_move_threshold_{ target_move_threshold }
+{
+  ros::NodeHandle pNh{ "~" };
+  target_pub_ = pNh.advertise<nav_msgs::Path>("smooth_control/target", 1);
+  closest_point_pub_ = pNh.advertise<nav_msgs::Path>("smooth_control/closest_point", 1);
+}
 
+RobotState SmoothControl::acquireNewTarget(const nav_msgs::PathConstPtr& path, const RobotState& state,
+                                           size_t state_index) const
```
Alescontrela: `state_index` goes from being called `path_index` to `path_index` again and then to `state_index`. Try to keep the same name for consistency.

igvc_navigation/src/path_follower/smooth_control.cpp
```diff
@@ -126,81 +127,129 @@ Action SmoothControl::getAction(const RobotState& state, const RobotState& targe
...
-  , simulation_frequency_{ simulation_frequency }
-  , target_velocity_{ target_velocity }
-  , lookahead_dist_{ lookahead_dist }
-  , simulation_horizon_{ simulation_horizon }
+size_t SmoothControl::getClosestIndex(const nav_msgs::PathConstPtr& path, const RobotState& state) const
 {
-  ros::NodeHandle pNh{ "~" };
-  target_pub_ = pNh.advertise<nav_msgs::Path>("smooth_control/target", 1);
-  closest_point_pub_ = pNh.advertise<nav_msgs::Path>("smooth_control/closest_point", 1);
+  auto closest = std::min_element(path->poses.begin(), path->poses.end(),
```
Alescontrela: very clean :+1:

oswinso:
Closing pull request to be merged together with motion profiling.

PR #374, opened 2019-03-24
Motion Profiling

oswinso:
This PR closes #374, changing the smooth control to take into account acceleration, so that the paths generated don't have an acceleration that goes beyond a limit. This is done by having desired velocity is a function of curvature (using the equation in the smooth control paper), so that during sections of high curvature the velocity is lowered to allow for the the robot to stay within acceleration limits while following the path.

Also, `PathFollower` is separated into `TrajectoryPlanner` and `TrajectoryFollower`, so that a controller can be added for following trajectories later on if desired.


igvc_gazebo/config/igvc_joint_control.yaml
```diff
@@ -7,8 +7,8 @@
   left_wheel_effort_controller:
     type: effort_controllers/JointEffortController
     joint: axle_to_left_wheel
-    pid: {p: 100.0, i: 0.01, d: 10.0}
+    pid: {p: 10000.0, i: 0.1, d: 100.0}
```
JasonGibson274: This doesn't actually do anything

igvc_description/urdf/jessi.urdf
```diff
@@ -42,7 +42,7 @@
     <inertial>
       <origin xyz="0 0 0" />
       <mass value="30.0" />
-      <inertia  ixx="0.994968" ixy="-0.100194"  ixz="0.0"  iyy="0.994968"  iyz="0.0"  izz="-1.0" />
+      <inertia  ixx="0.094969" ixy="-0.100194"  ixz="0.0"  iyy="0.094968"  iyz="0.0"  izz="-0.1" />
```
JasonGibson274: We should calculate these from the CAD for realsies.

oswinso:
Closing temporarily as I will be changing this to keep curvature constant.

igvc_description/urdf/jessi.urdf
oswinso: Reverted changes.

igvc_gazebo/config/igvc_joint_control.yaml
oswinso: Reverted changes.

igvc_navigation/src/trajectory_planner/motion_profiler.h
```diff
@@ -0,0 +1,46 @@
...
+struct RobotConstraint
+{
+  double velocity;
+  double acceleration;
+};
+
+struct MotionProfilerOptions
+{
+  double beta;
+  double lambda;
```
kylestach: These are not useful names. Please mention what they actually do, or, even better, change them to be actually descriptive. If you're going to keep them as beta/lambda, please reference some paper or other source which explains what they are in more detail (but prefer renaming, because it's a lot less overhead).

igvc_navigation/src/trajectory_planner/motion_profiler.cpp
```diff
@@ -0,0 +1,86 @@
...
+    {
+      double distance;
+      if (std::abs(trajectory_point.curvature) > 1e-8)
+      {
+        double R = 1 / trajectory_point.curvature;
+        double current_yaw = tf::getYaw(trajectory_point.pose.orientation);
+        double next_yaw = tf::getYaw(trajectory_ptr->trajectory[i + 1].pose.orientation);
+
+        double d_theta = next_yaw - current_yaw;
+        distance = std::abs(d_theta * R);
```
kylestach: This calculation should be very close to the below case for small delta yaw. Does it really need to be a separate case? (Hint: curvature was _probably_ calculated using the (dyaw)/sqrt(dx^2 + dy^2) formulation, so this really doesn't change anything).

If the answer to the above is yes, please actually mention what it does in a comment - i.e. "Approximate this segment of the path as an arc with radius `1 / curvature`".

igvc_navigation/src/trajectory_planner/motion_profiler.cpp
```diff
@@ -0,0 +1,86 @@
...
+          double capped_velocity =
+              trajectory_ptr->trajectory[i - 1].velocity + std::copysign(robot_constraint_.acceleration, dv) * dt;
+          ROS_WARN_STREAM_THROTTLE(1, "K: " << trajectory_point.curvature << ". Desired a: " << (dv / dt)
+                                            << ", limiting to " << robot_constraint_.acceleration
+                                            << ", desired velocity: " << trajectory_point.velocity << ", capping to "
+                                            << capped_velocity);
+          trajectory_point.velocity = capped_velocity;
+        }
+      }
+      ros::Duration move_duration = ros::Duration(distance / std::abs(trajectory_point.velocity));
```
kylestach: You actually really want the _average_ velocity over the entire segment here, right? i.e. you actually want to calculate _this_ segment's timestamp based on the distance from the previous segment to this segment and the average of the velocity at this segment and the previous one.

igvc_navigation/src/trajectory_planner/smooth_control.cpp
```diff
@@ -0,0 +1,211 @@
...
+  // Calculate timesteps
+  static ros::Duration dt = ros::Duration(1 / path_generation_options_.simulation_frequency);
+
+  // Visualization
+  nav_msgs::Path closest_point_path;
+  nav_msgs::Path targets_path;
+
+  closest_point_path.header = path->header;
+  targets_path.header = path->header;
+  for (int i = 0; i < path_generation_options_.getNumSamples() && path_index < path->poses.size() - 1; i++, time += dt)
```
kylestach:
```cpp
// Populate the trajectory with samples generated by the smooth control law
```
Also, it makes a lot more sense to calculate time in the loop with `ros::Time time = path->header.stamp + i * dt` instead of maintaining two separate iteration variables.

igvc_navigation/src/trajectory_planner/smooth_control.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+  // adjust both angles to lie between -PI and PI
+  igvc::fit_to_polar(delta);
+  igvc::fit_to_polar(theta);
+
+  // calculate the radius of curvature, K
+  double d = state.distTo(target);  // euclidian distance to target
+  double K = smooth_control_options_.k2 * (delta - atan(-smooth_control_options_.k1 * theta));
+  K += (1 + (smooth_control_options_.k1 / (1 + pow(smooth_control_options_.k1 * theta, 2)))) * sin(delta);
+  K /= -d;
```
kylestach: A few notes:
 - It doesn't make sense to mutate `K` here, the intermediaries have nothing to do with the curvature.
 - Use some actual intermediary equations here. Check the original paper (also probably cite it here) to see if they have any.
 - Cut down on visual noise: `const auto [k1, k2] = smooth_control_options_` already does a ton to clean this up, for example.

igvc_navigation/src/trajectory_planner/smooth_control.cpp
```diff
@@ -0,0 +1,211 @@
...
+size_t SmoothControl::getClosestIndex(const nav_msgs::PathConstPtr& path, const RobotState& state) const
+{
+  auto closest = std::min_element(path->poses.begin(), path->poses.end(),
+                                  [&state](geometry_msgs::PoseStamped lhs, geometry_msgs::PoseStamped rhs) {
+                                    return state.distTo(lhs.pose.position) < state.distTo(rhs.pose.position);
+                                  });
+  return static_cast<size_t>(closest - path->poses.begin());
+}
+
+double SmoothControl::blendCurvature(double distance, double K1, double K2) const
```
kylestach: This really shouldn't be necessary with a small enough dt, just as a heads up.

igvc_navigation/src/trajectory_planner/motion_profiler.cpp
```diff
@@ -0,0 +1,86 @@
...
+  , motion_profiler_options_{ motion_profiler_options }
+  , axle_length_{ axle_length }
+  , target_velocity_{ target_velocity }
+{
+}
+
+void MotionProfiler::profileTrajectory(const igvc_msgs::trajectoryPtr& trajectory_ptr, const RobotState& state)
+{
+  // For now, a simple "motion profile" where velocity is a function of curvature. ie. as curvature increases,
+  // velocity decreases.
```
kylestach: You're also profiling based on acceleration.

igvc_navigation/src/trajectory_planner/motion_profiler.cpp
```diff
@@ -0,0 +1,86 @@
...
+                               const MotionProfilerOptions& motion_profiler_options, double target_velocity)
+  : wheel_constraint_{ wheel_constraint }
+  , robot_constraint_{ robot_constraint }
+  , motion_profiler_options_{ motion_profiler_options }
+  , axle_length_{ axle_length }
+  , target_velocity_{ target_velocity }
+{
+}
+
+void MotionProfiler::profileTrajectory(const igvc_msgs::trajectoryPtr& trajectory_ptr, const RobotState& state)
```
kylestach: With how you currently are running things, you really don't gain anything by iterating over the entire trajectory. You only need to put a velocity profile over the segment between the current and next trajectory points, unless you're doing the backward pass (you're not).

igvc_navigation/src/trajectory_follower/trajectory_follower.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+TrajectoryFollower::TrajectoryFollower()
+{
+  ros::NodeHandle nh;
+  ros::NodeHandle pNh("~");
+
+  igvc::getParam(pNh, "topics/path", path_topic_);
+  igvc::param(pNh, "loop_hz", loop_hz_, 25.0);
+  igvc::param(pNh, "axle_length", axle_length_, 0.48);
+  igvc::param(pNh, "motor_loop_hz", motor_loop_hz_, 25.0);
```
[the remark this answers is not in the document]
oswinso: Yup. 😢

igvc_utils/src/state/robot_control.cpp
```diff
@@ -0,0 +1,33 @@
...
+  return velocity_pair;
+}
+
+CurvatureVelocity RobotControl::toKV(double axle_length) const
+{
+  if (axle_length == -1)
+  {
+    axle_length = axle_length_;
+  }
+  return { (right_ - left_) / axle_length, (right_ + left_) / 2 };
```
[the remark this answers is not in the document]
oswinso: Oh rip that's true.

igvc_navigation/src/trajectory_planner/smooth_control.h
```diff
@@ -0,0 +1,172 @@
...
+   */
+  size_t findTargetOnPath(const nav_msgs::PathConstPtr& path, const RobotState& target) const;
+
+  /**
+   * Finds the distance along the path provided from the current state to the target state
+   * @param path the path to calculate the distance along
+   * @param state the current state
+   * @param target the target state
+   * @return the distance from the current state to the target state along the given path
+   */
```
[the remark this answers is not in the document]
oswinso: yes, that's true.

igvc_navigation/src/trajectory_planner/motion_profiler.h
oswinso: yeah, they're in the Smooth Control paper. I'll link the paper in the comments.

igvc_navigation/src/trajectory_planner/smooth_control.cpp
oswinso: it's mentioned in the paper, but ehh you're probably right.

igvc_navigation/src/trajectory_planner/motion_profiler.cpp
oswinso: the original idea here was that it would be easier to implement something that did need to iterate over the entire trajectory in the future when we decide to make this better, but yeah you're right that iterating over the entire trajectory doesn't do anything here.

igvc_utils/src/state/robot_state.cpp
```diff
@@ -0,0 +1,209 @@
...
+  point.header.stamp = stamp;
+  point.pose = toPose();
+
+  const auto [k, v] = control.toKV(axle_length);
+  point.curvature = k;
+  point.velocity = v;
+  return point;
+}
+
+void RobotState::propogateState(const RobotControl &robot_control, double dt)
```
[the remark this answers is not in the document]
oswinso: heh thanks I can't spell.

igvc_navigation/src/trajectory_planner/smooth_control.cpp
oswinso: I agree that it doesn't make sense to mutate `K`, but like the [paper](https://web.eecs.umich.edu/~kuipers/papers/Park-icra-11.pdf) doesn't really have that many intermediary equations.

igvc_navigation/src/trajectory_planner/motion_profiler.cpp
oswinso: that's true.

igvc_navigation/src/trajectory_planner/motion_profiler.cpp
oswinso: yeah, this makes sense. Even though it doesn't _need_ to be a separate case, I think that we don't really lose anything by adding it in. I'll add a comment to clarify things

igvc_navigation/src/trajectory_planner/motion_profiler.cpp
oswinso: ah yes, that's true. I'll change it so it's more accurate.

igvc_navigation/src/trajectory_planner/smooth_control.cpp
oswinso: yeah, I'll move `time` into the loop body to make it clearer.

igvc_utils/include/igvc_utils/robot_control.h
```diff
@@ -19,7 +19,10 @@ class RobotControl
   friend std::ostream &operator<<(std::ostream &out, const RobotControl &robot_control);
 
   igvc_msgs::velocity_pair toMessage(ros::Time stamp) const;
-  CurvatureVelocity toKV(double axle_length = -1) const;
+  CurvatureVelocity toKV(const std::optional<double>& axle_length = std::nullopt) const;
+
+  double linearVelocity() const;
+  double angularVelocity(const std::optional<double>& axle_length = std::nullopt) const;
```
kylestach: Why would you ever need to call this with a null `axle_length`?

oswinso: I guess the thing I'm not sure about really is where `axle_length` should belong. It doesn't feel right that `axle_length` is a member of `RobotControl`, yet if it isn't then the caller needs to supply `axle_length` as an argument every time they want to convert from left / right velocities to curvature and linear velocity.

I guess the right thing to do here would be to just have the caller pass it every time, even though it's a pain to do.

PR #375, opened 2019-03-25
ROS Node and URDF updates for YostLabs IMU

Alescontrela:
This PR implements the functionality required to read linear acceleration, angular velocity, and orientation values from the IMU and publish these values to a ROS topic.

In short, the IMU is most easily calibrated via the [YostLabs 3-Space Sensor Suite](https://yostlabs.com/yost-labs-3-space-sensor-software-suite/) (Mac OSX, Windows). I'll write some documentation regarding the calibration of the IMU and i'll add it to the Wiki.


igvc_platform/launch/joystick_driver.launch
```diff
@@ -9,7 +9,7 @@
 <launch>
     <!-- Joystick node -->
     <node name="joy_node" pkg="joy" type="joy_node" output="screen" respawn="false">
-        <param name="dev" type="str" value="/dev/input/js0"/>
+        <param name="dev" type="str" value="/dev/igvc_joystick"/>
```
oswinso: Why are the joystick changes also in here?

Alescontrela: To prevent configuration conflicts with udev rules and to keep all our peripherals organized

oswinso: But this is a PR for the imu. I don't think it belongs in here.

Alescontrela: Sure, i'm going to make a separate PR

Alescontrela: Actually, `devel` contains the most recent udev rules. I just hadn't merged.

PR #379, opened 2019-03-29
Refactor waypoint_source to make use of the [utm->odom] transform supplied by navsat_transform_node

Alescontrela:
The previous implementation of `waypoint_source` made the assumption that `odom`'s x axis and `UTM`'s east axis were at a a 0 degree offset from one another, thus neglecting the robot's true heading as measured by the magnetometer. 

This PR corrects for this assumption by making direct use of the [`UTM`->`odom`] transform supplied by `navsat_transform_node`.


igvc_navigation/src/waypoint_source/waypoint_source.cpp
```diff
@@ -0,0 +1,149 @@
...
+void WaypointSource::position_callback(const nav_msgs::OdometryConstPtr& msg)
+{
+  geometry_msgs::PointStamped current_waypoint_odom;
+
+  // transform the waypoint at the head of the waypoints_ vector from
+  // UTM to odom
+  tf_listener_.transformPoint("odom", ros::Time(0), waypoints_.front(), "odom", current_waypoint_odom);
+
+  // move to the next waypoint in the list if we're < 1m away from the waypoint
+  if (igvc::get_distance(msg->pose.pose.position, current_waypoint_odom.point) < 1.0)
```
JasonGibson274: This should be a parameter

igvc_rviz_plugins/configs/jessii_conf.rviz
```diff
@@ -8,11 +8,13 @@ Panels:
         - /Status1
```
JasonGibson274: Some of these changes are arbitrary and should not be included.

JasonGibson274: Specifically the one that gives the rotation of the camera

igvc_navigation/src/waypoint_source/waypoint_source.cpp
Alescontrela: Will fix

Alescontrela: done :+1:

igvc_rviz_plugins/configs/jessii_conf.rviz
Alescontrela: I replaced the current config with the one from jessi, which is better organized and has a comprehensive list of the topics

PR #380, opened 2019-03-29
Fixed bug with path_follower getting stuck

### LINKED ISSUE #367: Path follower stops planning
During our testing day, we encountered a bug where the path follower becomes stuck and stops calculating trajectories for newer paths. I have a hunch that the node is stuck in an infinite loop or something similar, but neither the `path_follower` nor `smooth_control`nodes have loops that have the potential to become stuck.

See `jessii_field_with_lines_lowered_cost_harder_course_humans_last_2019-03-20-17-45-24.bag` on RJ Cloud for an example.

I have failed to reproduce the bug on both `devel` and  `feat/octomapper`.

oswinso:
Fixes #367. Changed fit_to_polar to use fmod and fixed get_closest_index.


Alescontrela:
LGTM :+1: Good fix

PR #381, opened 2019-04-02
removed unused serial implementation

JasonGibson274:
Removes the old serial implementation from 2015 since we no longer use it anymore.


Alescontrela:
LGTM :+1: run `clang-format` before merging though

PR #385, opened 2019-04-09
Fixed race condition with shared_ptr

oswinso:
This PR fixes the stupid non-deterministic race condition bug with the tests on `test_joystick_driver`.


Alescontrela:
:+1:

PR #397, opened 2019-04-19
Joystick control

oswinso:
This PR updates the joystick with a ncurses GUI, as well as adding a "direction-velocity" control that IMO feels more intuitive than tank controls. Also, there is some primitive acceleration bounding in order to prevent the accelerations when starting and stopping from being dangerously high.

This isn't fully complete, but I think there is enough changed here which warrants a PR, with issues + more PRs later down the road, given the importance (I think) of bounding acceleration now that our PID has feedforward.


oswinso:
ugh fine.

JasonGibson274:
Can I get the summary of why smooth control was put in the joystick? not how but why?

oswinso:
It was just a thing I wanted to test out. I still haven't found the time to get it working, so I've disabled it in the code.

JasonGibson274:
Okay why do we need it in the code base then?

oswinso:
We don't really _need_ it, just as how we don't _need_ another way of driving besides tank controls, but my idea was that it might be a more intuitive way to drive.

JasonGibson274:
I think setting u,w makes sense. I just don't understand the use case for smooth control.

igvc_platform/src/tests/CMakeLists.txt
```diff
@@ -1,2 +1,3 @@
 #add_rostest_gtest(TestJoystickDriver test/test_joystick_driver.test test_joystick_driver.cpp)
+#add_dependencies(TestJoystickDriver igvc_msgs_gencpp)
```
JasonGibson274: I thought this got fixed? am I wrong?

igvc_platform/src/joystick_driver/joystick_driver.h
```diff
@@ -0,0 +1,132 @@
...
+{
+  double left_x;
+  double left_y;
+  double right_x;
+  double right_y;
+};
+
+struct JoyMap
+{
+  int a = 0;
```
JasonGibson274: a_button, applies to all the ones below this

igvc_platform/src/tests/CMakeLists.txt
oswinso: I've changed the test so that it works for this implementation on my computer, but I still haven't figured out why it's not receiving messages on the CI server, which is why I commented it out.

igvc_platform/src/joystick_driver/joystick_driver.h
oswinso: Fixed

oswinso:
I'm taking inspiration from like rpg style movement, since I think that rpg movement feels pretty natural. I guess it's easier to explain once I fix it and can show it?

JasonGibson274:
I just don’t understand the use case. I would never want to drive like that. Now that is my opinion and I can very well be wrong.

JasonGibson274:
So I would say either give the solid use case or remove it. Your call.

oswinso:
My use case is that I can tell it where to go, and it will go in that direction vs telling it how much to turn. Either way though, I don't understand why a toggleable feature needs to be removed.

JasonGibson274:
Okay that’s your use case, that’s all I needed. When you are done doing me again and i’ll review.

JasonGibson274:
I think having a bunch of toggles where some work and some don’t should not be in the main code base. I should be able to toggle everything and it work as intended.

PR #352, opened 2019-02-11
Sim detector

kkeirstead3:
Implemented a color detector for the simulator that reads in the raw camera image and outputs a mono image with only the lines as white and everything else as black.

PR #399, opened 2019-04-20
Camera lidar projection

jviszlai:
Added a new node that subscribes to topics for a binary line detection image and a lidar pointcloud in order to project the lines into a 3d pointcloud. Pixels are projected onto a plane with a value of z that is calculated via interpolation between pixels. Initial z values are found by projecting lidar points into the camera coordinate space.

I also was using #352 to test the projection so maybe merge this after that one is merged.


igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+ros::Publisher _line_pointcloud_pub;
+ros::Subscriber image_sub;
+ros::Subscriber lidar_pointcloud_sub;
+sensor_msgs::CameraInfo::ConstPtr cam_info;
+cv::Mat cv_img;
+int resize_width;
+int resize_height;
+std::string line_topic;
+std::string pointcloud_topic;
+std::string lidar_topic;
```
oswinso: try to avoid globals if possible. For now though, at least make sure that they conform to ros c++ style guide, ie. prefix with g_.

http://wiki.ros.org/CppStyleGuide#Global_variables

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+ * Finds the two nearest neighbors to the input point, constraining the nearest neighbors
+ * to be from different lidar channels.
+ *
+ * @param point The point to find the nearest neighbors of
+ * @param kdtree_list List of 16 kdtrees, one for each lidar channel
+ * @param nearest1 The out parameter for the nearest pixel
+ * @param nearest2 the out parameter for the second nearest pixel from a different channel
+ */
+
+void find_nearest_neighbors(pcl::PointXY point, const std::vector<pcl::KdTreeFLANN<pcl::PointXY>> &kdtree_list,
```
oswinso: change to `findNearestNeighbors` to conform to ros c++ style guide, where camelCase is used for methods, and snake_case is used for variables.
http://wiki.ros.org/CppStyleGuide#Function_.2BAC8_Methods

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+                            pcl::PointXY &nearest1, pcl::PointXY &nearest2)
+{
+  float nearest1Dist = FLT_MAX;
```
oswinso: use snake_case for variables according to ros c++ style guide.

http://wiki.ros.org/CppStyleGuide#Variables

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+  float nearest2Dist = FLT_MAX;
+  for (auto kdtree : kdtree_list)
```
oswinso: change this to const reference if you're not mutating the kdtree

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+ *
+ * @param nearest1 1st nearest neighbor
+ * @param nearest2 2nd nearest neighbor
+ * @param point The point being interpolated
+ * @param xyz_img The 2D array that correlates each pixel to an x,y,z point
+ *
+ * @return Interpolated z value for point
+ */
+
+double interpolate_z(pcl::PointXY nearest1, pcl::PointXY nearest2, cv::Point2d point,
```
oswinso: same as above with method name.

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+#pragma omp parallel
+#pragma omp for collapse(2)
+  for (unsigned int r = 0; r < xyz_img.size(); r++)
+  {
+    for (unsigned int c = 0; c < xyz_img[0].size(); c++)
+    {
+      if (cv_img.at<uint8_t>(r, c) == 255 &&
+          !(std::find(mapped.begin(), mapped.end(), cv::Point2d(c, r)) != mapped.end()))
+      {
+        pcl::PointXY point = { (float)c, (float)r };
```
oswinso: prefer `static_cast` over c style casts.

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+        // Find ray, and intersection of ray at z = inter_z to determine x and y.
+
+        cv::Point3d cv_proj_ray = cam_model.projectPixelTo3dRay(cv_point);
+        tf::Vector3 proj_ray = tf::Vector3(cv_proj_ray.x, cv_proj_ray.y, cv_proj_ray.z);
+        tf::Matrix3x3 rotate_matrix = transform_cam_to_base.getBasis();
+        tf::Vector3 translation = transform_cam_to_base.getOrigin();
+        tf::Vector3 trans_ray = rotate_matrix * proj_ray;
+        double scale = (inter_z - translation.getZ()) / (trans_ray.getZ());
+        xyz_img[point.y][point.x].x = trans_ray.getX() * scale + translation.getX();
+        xyz_img[point.y][point.x].y = trans_ray.getY() * scale + translation.getY();
```
oswinso: consider refactoring this to a helper method.

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+  std::vector<cv::Point2d> mapped;
+  std::vector<pcl::PointCloud<pcl::PointXY>::Ptr> kdtree_cloud_list(16);
+  for (int i = 0; i < 16; i++)
+  {
+    kdtree_cloud_list[i] = std::unique_ptr<pcl::PointCloud<pcl::PointXY>>(new pcl::PointCloud<pcl::PointXY>);
+  }
+  std::vector<std::vector<pcl::PointXYZ>> xyz_img(cv_img.rows, std::vector<pcl::PointXYZ>(cv_img.cols));
+  pcl::PointCloud<pcl::PointXYZ> line_cloud;
+  // Calculate which lidar points are in the fov of the camera
+  for (auto point = msg->points.begin(); point != msg->points.end(); point++)
```
oswinso: again, use const reference if you're not mutating `point`

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+  {
+    tf::Vector3 lidar_point = tf::Vector3(point->x, point->y, point->z);
+    tf::Matrix3x3 rotate_matrix = transform_cam_to_base.getBasis();
+    tf::Vector3 translation = transform_lidar_to_cam.getOrigin();
+    tf::Vector3 transformed_point = rotate_matrix * lidar_point + translation;
+    cv::Point3d cv_point = cv::Point3d(transformed_point.getX(), transformed_point.getY(), transformed_point.getZ());
+    cv::Point2d pixel = cam_model.project3dToPixel(cv_point);
+    if (pixel.x >= 0 && pixel.x < cv_img.cols && pixel.y >= 0 && pixel.y < cv_img.rows)
+    {
+      mapped.push_back(pixel);
```
oswinso: prefer `emplace_back` over `push_back`

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+      int pixel_y = pixel.y;
+      int pixel_x = pixel.x;
+      tf::Matrix3x3 rotate_to_base = transform_lidar_to_base.getBasis();
+      tf::Vector3 translate_to_base = transform_lidar_to_base.getOrigin();
+      tf::Vector3 base_lidar_point = rotate_to_base * lidar_point + translate_to_base;
+      xyz_img[pixel_y][pixel_x] =
+          pcl::PointXYZ(base_lidar_point.getX(), base_lidar_point.getY(), base_lidar_point.getZ());
+      pcl::PointXY pcl_point = { (float)pixel_x, (float)pixel_y };
+      kdtree_cloud_list[point->ring]->push_back(pcl_point);
```
oswinso: same as above.

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+    }
```
oswinso: consider making this a helper method so it's easier to see what's happening.

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+  image_interpolation(xyz_img, mapped, kdtree_cloud_list);
+  for (unsigned int r = 0; r < xyz_img.size(); r++)
+  {
+    std::vector<pcl::PointXYZ> row = xyz_img[r];
+    for (unsigned int c = 0; c < row.size(); c++)
+    {
+      if (cv_img.at<uint8_t>(r, c) == 255)
+      {
+        pcl::PointXYZ point = row[c];
+        line_cloud.push_back(point);
```
oswinso: same as above.

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+    if (tf_listener.waitForTransform("/base_footprint", "/lidar", ros::Time(0), ros::Duration(3.0)))
+    {
+      tf_listener.lookupTransform("/base_footprint", "/lidar", ros::Time(0), transform_lidar_to_base);
+    }
+    else
+    {
+      ROS_ERROR_STREAM("\n\nfailed to find lidar to base footprint transform\n\n");
+    }
+  }
+  else
```
oswinso: consider putting this logic in a helper function to improve visual clarity.

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+/**
+ * Reads from velodyne pointcloud and uses a pinhole camera model in order to correlate pixels to 3d points.
+ * Publishes to a pointcloud topic, line_pointcloud, containing all 3d points that are part of a line.
+ */
+
+image_geometry::PinholeCameraModel cam_model;
+tf::StampedTransform transform_lidar_to_cam;
+tf::StampedTransform transform_cam_to_base;
+tf::StampedTransform transform_lidar_to_base;
+ros::Publisher _line_pointcloud_pub;
```
oswinso: try to avoid naming variables with prepended underscore. instead, do g_

igvc_perception/src/projection/main.cpp
jviszlai: I can't seem to find emplace_back and the compiler thinks I meant to type push_back.

igvc_perception/src/projection/main.cpp
jviszlai: I have `point++` which would prevent making it a const reference. Is there a better way to iterate through the pointcloud?

igvc_perception/src/projection/main.cpp
oswinso: that's strange. Uh... `mapped` is a `std::vector<cv::Point2d>`, which should have a `emplace_back` member function if you're working C++11 or newer. Double check everything I guess?

igvc_perception/src/projection/main.cpp
oswinso: you can use range-for loops, which is equivalent to this and something you used above as long as you're not doing things like finding indices by subtracting iterators or something.

`for (const auto& point : points)`

igvc_perception/src/projection/main.cpp
jviszlai: This case the `push_back` is for the `pcl::Pointcloud` type

igvc_perception/src/projection/main.cpp
jviszlai: It worked fine actually, nevermind.

igvc_perception/src/projection/main.cpp
oswinso: wait... I had no idea `pcl::PointCloud` had a `push_back` lol. I've been using `pcl::PointCloud.points.emplace_back` the entire time, since `pcl::PointCloud.points` is a `std::vector`. I guess this is fine then.

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,342 @@
...
+ * @param kdtree_list List of 16 kdtrees, one for each lidar channel
+ * @param nearest1 The out parameter for the nearest pixel
+ * @param nearest2 the out parameter for the second nearest pixel from a different channel
+ */
+
+void findNearestNeighbors(pcl::PointXY point, const std::vector<pcl::KdTreeFLANN<pcl::PointXY>> &kdtree_list,
+                          pcl::PointXY &nearest1, pcl::PointXY &nearest2)
+{
+  float nearest_1_dist = FLT_MAX;
+  float nearest_2_dist = FLT_MAX;
```
oswinso: didn't catch this before, but I think using `std::numeric_limits<float>::max()` is more readable here.

igvc_perception/src/projection/main.cpp
```diff
@@ -0,0 +1,342 @@
...
+                                       g_cam_info->P[6] * haf,
+                                       0.,
+                                       0.,
+                                       0.,
+                                       1.,
+                                       0. };
+    new_g_cam_info.K = new_K;
+    new_g_cam_info.P = new_P;
+    g_cam_model = image_geometry::PinholeCameraModel();
+    g_cam_model.fromCameraInfo(new_g_cam_info);
```
oswinso: you could probably also split this off

Issue #413, opened 2019-04-26
tf tree loop error caused by conflicting utm->odom publishers


Alescontrela:
When launching `qualification.launch` and `navsat_transform_node` (with `broadcast_utm_transform = true`) within the same roscore, a loop occurs in the tf tree due to conflicting utm->odom publishers.

steps to reproduce:
```bash
$ roslaunch igvc_gazebo qualification.launch
$ roslaunch igvc_navigation localization.launch
$ rosrun tf tf_echo base_footprint odom # spits out error
```

hot fix: 
not launching `ground_truth_publisher` results in the error not occurring, but in practice the `ground_truth_publisher` should stop publishing its transform once `navsat_transform_node` publishes its own.

PR #414, opened 2019-04-27
Changes odom to utm order in ground_truth_publisher

oswinso:
This PR swaps the parent and child frames in ground_truth_publisher, as well as making it stop publishing when another transform from odom to utm is found, closing #413.


igvc_gazebo/nodes/ground_truth/main.cpp
```diff
@@ -95,14 +125,7 @@ int main(int argc, char** argv)
...
-  tf::TransformBroadcaster br;
 
-  ros::Rate rate(20);
-  while (ros::ok())
-  {
-    br.sendTransform(tf::StampedTransform(utm_to_odom, ros::Time::now(), "utm", "odom"));
-    ros::spinOnce();
-    rate.sleep();
-  }
+  ros::Timer utm_timer = nh.createTimer(ros::Duration(5.0), boost::bind(utm_callback, _1, utm_to_odom.inverse()));
```
Alescontrela: make this duration less. Otherwise simulation takes 5 seconds to load the initial waypoints file.

oswinso: Changed to 1 second.

PR #419, opened 2019-04-30
Commit streaming slot settings to resolve IMU value-ordering issue

### LINKED ISSUE #418: IMU value ordering issue while field testing
On the most recent field testing day, we noticed that the imu and magnetometer values for heading and angular velocity were flipped. The most likely cause of this bug is that, when calibrating the imu with the [Yost Labs 3-Space Sensor Software Suite](https://yostlabs.com/yost-labs-3-space-sensor-software-suite/), the streaming slot order for the imu is changed from the expected streaming slot order on the ROS node.

Alescontrela:
This PR solves the issue encountered during our last field testing day whereby angular velocity and heading values were flipped. Streaming slots are now correctly signaled, ensuring proper access of IMU data. Closes #418

### DIFF
--- a/igvc_platform/launch/imu_top.launch
+++ b/igvc_platform/launch/imu_top.launch
@@ -6,7 +6,7 @@
     <!-- Top IMU -->
     <node pkg="igvc_platform" name="imu_top" type="imu" output="screen" >
         <param name="SERIAL_PORT" type="string" value="/dev/imu_top" />
-        <param name="frame_id" type="string" value="magnetometer"/>
+        <param name="frame_id" type="string" value="imu"/>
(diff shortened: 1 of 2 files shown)


igvc_platform/launch/imu_top.launch
```diff
@@ -6,7 +6,7 @@
     <!-- Top IMU -->
     <node pkg="igvc_platform" name="imu_top" type="imu" output="screen" >
         <param name="SERIAL_PORT" type="string" value="/dev/imu_top" />
-        <param name="frame_id" type="string" value="magnetometer"/>
+        <param name="frame_id" type="string" value="imu"/>
```
oswinso: What's this change for?

PR #422, opened 2019-05-01
rviz IMU plugin

oswinso:
This PR creates a new rviz plugin that visualizes the accelerometer, gyroscope and orientations. Right now, the class is a bit messy, since the coding style in the tutorial sucks, but in the future, we could probably create better helper classes than the ones provided in order to make the code much more cleaner.

[Example of plugin](https://www.youtube.com/watch?v=x1i0xnmZJRo&feature=youtu.be)


kylestach:
Why not something off-the-shelf like http://wiki.ros.org/rviz_imu_plugin?

oswinso:
I tried it out but there was no option for removing the gravity vector from the accelerometer && no gyro viz

kylestach:
ACK, fair enough.

Issue #479, opened 2019-09-04
Migrate navigation stack to move base flex


oswinso:
After some discussions (and making another decision then coming back) (see [discussion doc](https://docs.google.com/document/d/1Cv4E_ZGLk6-9t5osY7OML0yQu_HJbjkMvUyK5TOqvN8/edit) with some summaries of different approaches).

In short, the reasons why we are going with [move base flex](https://github.com/magazino/move_base_flex/) instead of porting [teb local planner](https://github.com/rst-tu-dortmund/teb_local_planner) to our custom navigation stack:
* move base flex fixes a lot of the problems issues that move base has
    * move base has a fixed state machine that only runs the global planner once, see [here](https://github.com/ros-planning/navigation/blob/d60209b348add7db379a50aaac3de8af3c7716e7/move_base/src/move_base.cpp#L859-L992)
    * move base locks the map representation to use costmap2d
* teb local planner is coupled with `costmap2d`, see [here](https://github.com/rst-tu-dortmund/teb_local_planner/blob/886ffc8587262cf0df7b559b116a75d1af6ad16b/src/teb_local_planner_ros.cpp#L369-L383). Using teb local planner would require porting to `costmap2d`.
* Many of the refactors that we would have done on our current navigation stack mirror the changes that move base flex brings to move base, ie. moving everything to action server based to free up changes to the state machine
* Moving to move base flex allows us to more easily move to other controllers in the long term

As discussed during the meeting:
* **global planner**
    * Encode "correct direction to move", should be very sparse
    * Only navigate using lines, no barrels
* **local planner**
    * Navigates using a _sparse_ global plan
    * Navigates on map with both lines and barrels

(tentative) plan for performing this refactor:

#### Stage 1: "Move" all parts under move base flex framework without actually refactoring
Move the global planner, local planner and mapper to use the move base flex framework, but without actually using the interfaces correctly
ie. Override the constructor, spin a separate thread that does all the original work.

The idea is to get the move base flex framework running, even if it isn't doing anything yet.

#### Stage 2: Actually change the parts to use the correct move base flex interface methods
Create "state machine" node that for now just forwards everything
Change each to use the correct move base flex interface methods
At the same time, change the "state machine" node to now use the correct actionlib servers from movebase

oswinso:
I think with #535 this epic should be completed.

PR #488, opened 2019-09-10
Created costmap2d layer thin wrapper around current mapper

### LINKED ISSUE #485: Move current mapping stack to costmap2d layers as thin wrapper
Per the refactoring plan in #479, the first step will be to move all the existing code to use costmap2d as only a "thin wrapper" by spinning up a separate thread, so that we are technically running our own stuff, but with costmap2d as a requirement.

AC: The current mapper, but as a separate thread that is spun off a costmap2d layer.

oswinso:
# Description

With our refactor to move base flex, we need to refactor our mapper to use `costmap2d`.
As laid out in #479, the first step of the refactoring is to move all the current nodes (mostly unchanged) to "use" the equivalent move base flex interfaces without actually implementing them properly.

This PR does the following:
- Moves the `main` method out of `ros_mapper.cpp` and into a `mapper_node.cpp`
- Creates a `mapper_lib` library
- Creates a `WrapperLayer` that extends `costmap_2d::Layer` and instantiates a `ROSMapper` (so the old functionality is unchanged)
- Creates a `mapper_node.cpp` that includes a main file which instantiates `ROSMapper` (in case we find a bug in `WrapperLayer` and want to run it by itself)
- Creates a `config` folder in `igvc_navigation` that contains all the config for move base flex
    - Additionally moves `octomap.yaml` and `octomap_sim.yaml` into this folder
- Creates a `mapper_wrapper.launch` file that launches move base flex, although only the global map part of this stack is useful.
- Creates config `.yaml` files for all the components for move base flex (these do nothing right now except for `global_costmap_params.yaml`

Closes #485 

# Testing steps (If relevant)
## Test that the old functionality still works in simulation
1. Run simulation by launching `igvc_gazebo qualification.launch`
2. Run the navigation stack by launching `igvc_navigation navigation_simulation.launch`

Expectation: The mapper should still work and the robot should still be able to map the environments, plan a path and follow that path. The `mapper/debug/pcl` topic should show the map.

# Self Checklist


Alescontrela:
You might need to add `move_base_flex` to the `packages.xml`

PR #495, opened 2019-09-11
Feat/implement costmap2d interfaces

### LINKED ISSUE #486: Change the "thin wrapper" costmap2d layer to actually use the correct costmap2d interfaces
After #485, we will have moved our mapping system to use `costmap2d` as a "thin layer".

The task now is to actually use the interfaces provided by `costmap2d` so that we are using `costmap2d` properly. For backwards compatibility reasons (to aid refactoring of the planners) both the old interface (separate /map topic with custom type) and new `costmap2d` interfaces will be supported.

AC: `costmap2d` layer interfaces implemented, but keeping backwards compatibility.

oswinso:
# Description

Per the refactor plan, the next step is to actually implement the interfaces for costmap2d

This PR does the following:
- Implement the interfaces for the costmap2d layer for our mapper

Fixes #486

# Testing steps (If relevant)
## Test that the old functionality still works in simulation and the costmap2d interfaces are implemented
1. Run simulation by launching `igvc_gazebo qualification.launch`
2. Run the navigation stack by launching `igvc_navigation navigation_simulation.launch`

Expectation: There should be a topic for the global costmap that shows the map in rviz.

# Self Checklist

### DIFF
--- a/igvc_navigation/config/global_costmap_params.yaml
+++ b/igvc_navigation/config/global_costmap_params.yaml
@@ -2,8 +2,17 @@ global_costmap:
     global_frame: odom
     plugins:
         - {name: wrapper_layer,     type: "wrapper_layer::WrapperLayer"}
-    publish_frequency: 1.0
+        - {name: inflation_layer,   type: "costmap_2d::InflationLayer"}
+    publish_frequency: 5.0
     robot_radius: 1.0
     width: 200
     height: 200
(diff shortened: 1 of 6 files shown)


oswinso:
Force pushed to rebase onto master

oswinso:
Ive created an issue #498 for integrating lines into sim.

Issue #487, opened 2019-09-09
Refactor mapper according to design document


oswinso:
With #486 done and the `costmap2d` layer API implemented, the next step will be to actually perform the refactor following the design document laid out in #484.

The old interface will be removed, so moving the planners to use the move base flex interfaces will need to be done before this is merged in.

AC: Two separate costmap layers that perform occupancy grid mapping, with the old interfaces removed.

oswinso:
I've been working on this, and I feel like this should be separated into the lidar and line part separately to make the PRs smaller (and easier to review).

Issue #474, opened 2019-08-25
Barrel segmentation


oswinso:
With a 3D lidar, we need to be able to be able to detect and segment barrels from the 3D pointcloud.

Our current approach is to segment the ground plane instead, with the residual points being counted as a "barrel". However, this method is not robust against a ground that is not flat.

PR #509, opened 2019-09-21
Pointcloud filter node

### LINKED ISSUE #505: Pointcloud Filter nodelet
Per the [design document on the mapper refactor](https://github.com/RoboJackets/igvc-software/blob/master/documents/design/mapping_refactor.md), we will be splitting off the multiple responsibilities that the current mapper has into individual node/nodeletts. One of them is the pointcloud filtering.

AC: Node & Nodelett that perform the same pointcloud filtering that the current mapper does right now, ie. start angle - end angle, empty space

oswinso:
# Description

With the mapper refactor to move base flex #479 and #487, the first thing we need to do is to refactor the "pointcloud filtering" responsibilities of the current mapper to a new node.

This PR does the following:
- Creates a `pointcloud_filter` node to perform pointcloud filtering
    - This is not a nodelett as I discovered that as long as the publisher and subscriber are
    from the same process, publishing using `boost::shared_ptr` will result in no copy.
- Creates filters that mimic the existing mapper's pointcloud filtering capabilities (with the exception of the ground plane filter since #474 will implement a better one
- Creates a `pointcloud_filter.launch` inside `igvc_perception` that launches this node.

Fixes #505.

# Testing steps (If relevant)
## Pointcloud filter works
1. Launch simulation with `roslaunch igvc_gazebo qualification.launch`
2. Launch the pointcloud filter node with `rosluanch igvc_perception pointcloud_filter.launch`
3. (Optional) Launch the navigation stack with `roslaunch igvc_navigation navigation_simulation.launch`
4. (Optional) Move barrels around in gazebo

Expectation: The topics `lidar/occupied` and `lidar/free` yield the occupied points and endpoints of the free space effectively.

# Self Checklist


oswinso:
Great! Thanks for the review. Merging.

Issue #496, opened 2019-09-12
Add teb local planner to our stack


oswinso:
As a part of the path planning refactor, we are going to try out [teb local planner](http://wiki.ros.org/teb_local_planner) for our local path planning.

To do that, we have started refactoring our current stack on to move base flex in #479.

However, to test `teb local planner`, we only really need to to have the mapper refactored, which will have happened in #495. Thus, once that happens, we should be able to try out `teb local planner` on our stack.

To get started, check out the [teb's tutorials on the ros wiki](http://wiki.ros.org/teb_local_planner/Tutorials/Configure%20and%20run%20Robot%20Navigation).

Since teb local planner uses the local costmap, but only the global costmap has the wrapper layer, you will need to configure the local costmap to use a `static layer` that takes in the global costmap's map as input. Check [this](http://wiki.ros.org/costmap_2d/hydro/staticmap) for the parameters that `static layer` takes to set that up.

Also, since the `move base flex` framework outputs `geometry_msgs/Twist`, we need to convert that to the `igvc_msgs/VelocityPair` that gets sent on the `motors` topic for the `motor_controller` node to send that down to the mbed.

The kinematics can be found in google by googling "Differential Drive Kinematics"
For example, [here](http://www.cs.columbia.edu/~allen/F17/NOTES/icckinematics.pdf)

Acceptance Criteria: Jessii moving in simulation with Teb Local Planner.

PR #514, opened 2019-10-02
`lidar_layer` part of the mapper refactor

### LINKED ISSUE #513: Create `lidar_layer`
The mapper refactor design document outlines the creation of a `lidar_layer` layer for `costmap_2d` that extends `costmap_2d::Layer`, which performs the lidar part of the map.

AC: A `lidar_layer` plugin that extends `costmap_2d::Layer` and performs the same lidar mapping functionality as the old mapper.

oswinso:
# Description
With the refactor to move base flex, we needed to move our mapping stack to `costmap_2d`. The design document for this refactor was outlined in #483, and the issue for the refactor in #487. However, while writing the first `lidar_layer`, I realized that this would be quite a large PR, and so I decided to split the issue into the two parts: `lidar_layer` and `line_layer`.

While I originally decided to write my own array for the backing structure, I came across [gridmap](https://github.com/ANYbotics/grid_map), which seems to be a nice library to use for 2d grid maps. As a result, I have added `grid_map` as a dependency and used that instead. 

To keep maintain compatibility, a `refactored_mapper.launch` launch file has been created that  launches the new `lidar_layer` instead of the old `wrapper_layer`. Since the path planner hasn't been migrated yet (rip issues) and #496 hasn't been merged in yet, we can't actually run the full navigation stuck using this new `lidar_layer` yet.

This PR does the following:
- Creates a `line_layer` `costmap_2d::Layer` that follows the design document
- Create a new `refactored_local_costmap_params.yaml` file that loads the new `line_layer` as a local planner and has some params.
- Adds `grid_map` as a dependency to our stack

Fixes #513

# Testing steps (If relevant)
## New `line_layer` does as expected
1. Run simulation. `roslaunch igvc_gazebo qualification.launch`. Add a few barrels in front of the robot
2. Run the new `lidar_layer`. `roslaunch igvc_navigation refactored_mapper.launch`
3. Open up rviz and add a display of type `Map` for the topic `/move_base_flex/local_costmap/costmap`.
4. Optional: Since the `costmap` viz is pretty bad and discretizes everything, to view the undescritized probabilities, do `roslaunch igvc_navigation gridmap_viz.launch` and add a display of type `Map` for the topic `/grid_map_visualization/occupancy_grid`. You should see a (slightly less crappy) visualization that properly shows the 0.0 - 1.0 occupancy probabilities (although its hard to judge where 0.5 is).

Expectation: The places where the barrels are marked as occupied in the map (whichever one you choose to use). If you take the barrels away, then occupancy mapping happens and after a while the place is marked as free space.

Note: The above behaviour only really happens because the occupancy probabilities are capped at 0.01 and 0.99. This is located in the `refactored_local_costmap_params.yaml` file.

Also, please review this PR locally, as imo it's kind of hard to review large changes on github.

# Self Checklist

### DIFF
--- a/igvc_navigation/CMakeLists.txt
+++ b/igvc_navigation/CMakeLists.txt
@@ -25,6 +25,15 @@ find_package(catkin REQUIRED COMPONENTS
              costmap_2d
              dynamic_reconfigure
              igvc_utils
+             grid_map_core
+             grid_map_ros
+             grid_map_cv
+             grid_map_filters
+             grid_map_loader
+             grid_map_msgs
+             grid_map_octomap
+             grid_map_rviz_plugin
+             grid_map_visualization
     )
 
 set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
@@ -190,7 +199,8 @@ add_library(
     GraphSearch STATIC
     GraphSearch/src/Node.cpp
     GraphSearch/src/Graph.cpp
-    GraphSearch/src/PriorityQueue.cpp)
+    GraphSearch/src/PriorityQueue.cpp
+)
 add_dependencies(GraphSearch ${catkin_EXPORTED_TARGETS})
 # link OpenCV and cv_bridge to GraphSearch library
 target_link_libraries(GraphSearch ${OpenCV_LIBRARIES} ${catkin_LIBRARIES})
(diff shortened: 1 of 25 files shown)


tangemicioglu:
Looks good and works, don't forget to add issues for the TODOs and eventually getting rid of wrapper layer.

PR #515, opened 2019-10-02
Changes camera frame + topics to use namespaces more, clean up segmented image and camera_info topic, adds camera intrinsics matrix scaling and publishing for sim_detector

### LINKED ISSUE #450: Standardize segmented image and camera frame names
Right now, the topic name of the segmented images and the names of the camera frames for simulation and real life are different. Change the name of the camera frame and topic names in simulation to match the one in real life.

AC: Camera frame and image topics for both real life and simulation are the same.

oswinso:
# Description

Even though #503 was merged which closed #450, I realized that the camera frame and topic names were still kind of not that nice. What I mean by this is that the `sensor_msgs::CameraInfo` and `sensor_msgs::Image` topics for a camera are like `cam/center/CameraInfo` and `cam/center/image_raw`.

However, for the output of the neural network, which scales the image (and thus needs a different `sensor_msgs::CameraInfo`, the the `sensor_msgs::Image` topic is `cam/center/semantic_segmentation`. The problem thus comes with where the `sensor_msgs::CameraInfo` for the segmented image should go. According to the `image_transport::CameraSubscriber`, and `image_transport::ImageTransport`, the topic for the `sensor_msgs::CameraInfo` should be in the same namespace as the `sensor_msgs::Image` topic, which makes things quite awkward for the segmented image as that would result in the same topic as the raw image.

In order to solve this problem, and also make better use of namespaces, this PR does the following: 
- Rename the "raw" camera image topics to `cam/center/raw/image`, with the `CameraInfo` topic being `cam/center/raw/camera_info`
- Rename the segmented image topic to `cam/center/segmented/image`, with the corresponding `CameraInfo` topic being `cam/center/segmented/camera_info`
- Add a `scaleCameraInfo` method to `sim_color_detector` so that it correctly publishes both the 400x400 segmented image, as well as the modified `CameraInfo` of the segmented image
- Change the _old mapper_ to use the new camera topics by updating the topic names in `octomap_sim.yaml` and `octomap.yaml`
- Update the topics in `camera.launch`

Fixes #450 again.

Note: Neither `cnn.py` nor `vision.launch` was updated. This is because I think that both files need to be refactored (multiple responsibilities, make the file cleaner etc.), and so I will make a new issue for that specifically.

# Testing steps (If relevant)
## Mapper still works with the updated frames
2. Launch old mapper: `roslaunch igvc_navigation navigation_simulation.launch`
3. Open up the `mapper/debug/map_debug_pcl` topic on rviz

Expectation: Same behavior as before. Mapper is able to subscribe to segmented images and lines are visible on the map.

# Self Checklist

### DIFF
--- a/igvc_description/urdf/jessii.urdf.xacro
+++ b/igvc_description/urdf/jessii.urdf.xacro
@@ -257,7 +257,7 @@
   </xacro:macro>
 
 
-  <link name="center_cam">
+  <link name="cam/center">
       <visual>
           <geometry>
               <box size="0.1 0.1 0.1"/>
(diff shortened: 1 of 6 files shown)


oswinso:
Cool, thanks for the review! Merging.

PR #491, opened 2019-09-11
cmake git submodule check, added common problems to readme

oswinso:
# Description

It is very hard to notice that git submodules are missing. When one tries to compile without the `rj-ros-common` git submodule, one gets the error message:
`Could not find a package configuration file provided by "parameter_assertions" with any of the following names` which is not helpful at all in diagnosing the problem.

This PR does the following:
- Adds a `ensure_submodules.cmake` script to cmake which will automatically call `git submodule update --init --recursive` to ensure that `rj-ros-common` is present during a build
- Adds a "common problems" section to the readme, which includes the git submodule problem (although this shouldn't happen anymore)

Fixes #490, fixes #453

# Testing steps (If relevant)
## cmake automatically does `git submodule update --init --recursive`
1. Remove the `external/rj-ros-common` folder with `rm -rf external/rj-ros-common`
2. Remove the build and devel folders (otherwise cmake caching is gonna complain) `rm -rf build devel`
3. Run `catkin_make`

Expectation: Somewhere in the log you'll see a
```
-- Submodule update
Submodule path 'external/rj-ros-common': checked out '4578f74ebf294a470952ce190e3299176f6bdd86'
```
and the `external/rj-ros-common` folder should be present.

# Self Checklist
- [x] I have formatted my code using `make format`
- [x] I have tested that the new behaviour works

PR #517, opened 2019-10-03
Changed ensure_submodules to check for .git folder inside submodule

### LINKED ISSUE #490: Hard to tell that git submodules are not initialized
Because of how dumb git submodules are, it is almost impossible to tell that you are missing git submodules if you are not already aware of them.

AC: Some system to allow people who do a regular `git clone` to be able to solve the problem of missing the `rj-ros-common` git submodule.

### LINKED ISSUE #516: rj-ros-common check doesn't work as intended
When cloning the repository, an empty rj-ros-common folder is created under external, which doesn't trigger the check from #491

oswinso:
# Description

I thought we fixed #490 with #491, but turns out the submodule folder _does_ exist, it's just that there's nothing inside it.

This PR does the following:
- Change the check to check the `.git` folder inside the submodule instead of the folder itself

Fixes #516.

# Testing steps (If relevant)
## Check works for if the submodule folder exists but is empty
1. Delete `igvc-software` folder
2. Clone it without `--recursive`
3. Run `catkin_make`

Expectation: It should scream at you telling you to run `catkin_make` again.

# Self Checklist

### DIFF
--- a/cmake/ensure_submodules.cmake
+++ b/cmake/ensure_submodules.cmake
@@ -29,9 +29,9 @@ if(NOT EXISTS ${GIT_REPO_DIR}/external)
 endif()
 
 foreach(EXTERNAL_PACKAGE ${EXTERNAL_PACKAGES})
-    if(NOT EXISTS ${GIT_REPO_DIR}/external/${EXTERNAL_PACKAGE})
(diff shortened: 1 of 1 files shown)


oswinso:
Cool, I'm merging.

PR #518, opened 2019-10-09
Adds TEB Local Planner and launching navigation from move base flex

### LINKED ISSUE #496: Add teb local planner to our stack

tangemicioglu:
# Description

One of the main goals of moving to move base flex as mentioned in #479 is to be able to use the TEB local planner. This PR integrates TEB into our stack and creates a way to launch navigation using it.

This PR does the following:
- Adds TEB local planner to move base flex
- Establishes interfaces between TEB and other parts of stack
- Creates a way of launching TEB with a basic configuration

Fixes #496

# Testing steps (If relevant)
## Make sure TEB actually works and the old local planner is also functional
1. Launch simulation with `roslaunch igvc_gazebo qualification.launch`
2. Launch old local planner with `roslaunch igvc_navigation navigation_simulation.launch` and run it until you're convinced it works
4. Repeat 1, launch TEB with `roslaunch igvc_navigation navigation_simulation_teb.launch` and let it move until it reaches both waypoints.

Expectation: The old local planner should work the same way it used to. TEB should be able to reach the waypoints.

# Self Checklist
- [X] I have tested that the new behavior works


igvc_navigation/src/differential_drive/differential_drive.cpp
```diff
@@ -0,0 +1,44 @@
...
+{
+  ros::NodeHandle nh;
+  ros::NodeHandle pNh("~");
+
+  assertions::param(pNh, "axle_length", axle_length_, 0.48);
+  assertions::param(pNh, "max_vel", max_vel_, 3.0);
+  ros::Subscriber mbf_twist = nh.subscribe("/cmd_vel", 1, &DifferentialDrive::twistToVelocity, this);
+  vel_pub_ = nh.advertise<igvc_msgs::velocity_pair>("/motors", 1);
+
+  ros::spin();
```
oswinso: Also, (this is something that should also be changed in a few other nodes), it's better to have the `ros::spin()` in the main than to have it in the constructor, since theoretically this would allow us to have multiple nodes running in an executable by constructing them in `main`.

igvc_navigation/src/mapper/wrapper_layer.cpp
```diff
@@ -32,6 +32,7 @@ void WrapperLayer::updateCosts(costmap_2d::Costmap2D& master_grid, int min_i, in
   std::optional<cv::Mat> map = mapper_.mapper_->getMap();
   if (!map)
   {
+    current_ = false;
```
oswinso: Leftover from debugging?

tangemicioglu: This was intentional, because MBF (or the planners, I don't remember exactly) decides if it can or can't use a map based on whether it's current.

igvc_navigation/src/action_server/action_server.cpp
```diff
@@ -0,0 +1,54 @@
...
+  mbf_msgs::MoveBaseGoal goal;
+
+  goal.target_pose.header = point.header;
+  goal.target_pose.pose.position = point.point;
+
+  tf::Quaternion quat_tf;
+  geometry_msgs::Quaternion quat_msg;
+  quat_tf.setRPY(0, 0, 1);
+  tf::quaternionTFToMsg(quat_tf, quat_msg);
+  goal.target_pose.pose.orientation = quat_msg;
```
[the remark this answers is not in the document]
tangemicioglu: I knew there was something for this, but just couldn't find it. Thanks!

Issue #532, opened 2019-10-13
Teb doesn't publish anything


tangemicioglu:
Teb broke again after some recent modifications, and needs to be fixed for the new navigation to work.
AC: Teb can navigate around the track again.

PR #524, opened 2019-10-10
Feat/line layer

oswinso:
# Description
Note: #515 should be merged in before this as this relies of the changes in #515.

This PR does the following:
- Changes `sim_detector` to work in HSV
- Changed mimap bias for the  "map course" texture in gazebo to reduce the "blurriness" when lookng at lines at an angle, messing up the line detector
- Implemented the line layer functionality

Fixes #523.

# Testing steps (If relevant)
## Lines work
1. Launch simulation. `roslaunch igvc_gazebo qualification.launch`
2. Run new mapping stack. `roslaunch igvc_navigation refactored_mapper.launch`
3. Open `rviz` and have `move_base_flex/local_costmap/costmap` open as a display

Expectation: Lines show up on the map.

## Regression: Barrels still work
1. Launch simulation. `roslaunch igvc_gazebo qualification.launch`
2. Run new mapping stack. `roslaunch igvc_navigation refactored_mapper.launch`
3. Open `rviz` and have `move_base_flex/local_costmap/costmap` open as a display
4. Move a barrel in front of the robot

Expectation: Both lines and barrels show up on the map.

# Self Checklist
- [x] I have formatted my code using `make format`
- [x] I have tested that the new behaviour works

PR #535, opened 2019-10-13
Final mapper + TEB integration as part of MBF refactor

### LINKED ISSUE #521: Deprecate WrapperLayer and old navigation stack
As soon as the line layer PR gets merged (TODO) and #520 gets merged (giving us the same functionality as before the refactor to move base flex), we should be able to deprecate the WrapperLayer and the old navigaiton stack

AC: Make `lidar_layer` and `line_layer` the default and deprecate `wrapper_layer` and the old nvaigation stack for `navigation_simulation.launch`. Also potentially reorganize the `igvc_navigation` package, and split the old / new navigation stack.

### LINKED ISSUE #542: Make TEB stop caring about waypoint orientations
The current "dummy orientation" solution in the `action_server` will result in weird trajectories when the path is facing a different direction. A simple config change should fix that.

oswinso:
# Description

Finally, both the mapper refactor and the move base flex integration has been completed in #524 and #518. We should now be able to move off of the old navigation stack to the new stack after the current TEB bug on master, #532, is fixed (assuming nothing breaks).

This PR does the following:
- Changes `navigation_simulation.launch` to launch the new mbf stack instead of the old one
- Adds a publisher to the `line_layer` to publish it's own map
- Changes the global map to be a static layer that just takes in the line layer's map

Fixes #521, fixes #542 

# Testing steps (If relevant)
## [Regression] The entire navigation stack still works
2. Launch the navigation stack. `roslaunch igvc_navigation navigation_simulation.launch`
3. Open up `rviz`

Expectation: We can navigation from start to goal

# Self Checklist


igvc_navigation/config/teb_local_planner_params.yaml
```diff
@@ -73,14 +79,14 @@ TebLocalPlannerROS:
...
-  weight_viapoint: 1
+  weight_viapoint: 0.0001
   weight_adapt_factor: 2
 
   # Homotopy Class Planner
 
   enable_homotopy_class_planning: False
 #  enable_multithreading: True
-#  max_number_classes: 4
+#  max_number_classes: 2
```
tangemicioglu: Not sure what the point of this was, but we'll probably enable this after #538 is resolved.

igvc_navigation/config/global_costmap_params.yaml
```diff
@@ -2,8 +2,8 @@ global_costmap:
     global_frame: odom
     robot_base_frame: base_link
     plugins:
-        - {name: wrapper_layer,     type: "wrapper_layer::WrapperLayer"}
```
tangemicioglu: Can't we remove the files for  `wrapper_layer` at this point?

igvc_navigation/launch/refactored_mapper.launch
```diff
@@ -1,13 +1,13 @@
 <launch>
     <!-- Move Base Flex -->
-    <node pkg="mbf_costmap_nav" type="mbf_costmap_nav" respawn="false" name="move_base_flex" output="screen">libsimple_layer
+    <node pkg="mbf_costmap_nav" type="mbf_costmap_nav" respawn="false" name="move_base_flex" output="screen" >
         <param name="tf_timeout" value="1.5"/>
         <param name="planner_max_retries" value="3"/>
         <rosparam file="$(find igvc_navigation)/config/planners.yaml" command="load" />
         <rosparam file="$(find igvc_navigation)/config/controllers.yaml" command="load" />
         <rosparam file="$(find igvc_navigation)/config/recovery_behaviors.yaml" command="load" />
-        <rosparam file="$(find igvc_navigation)/config/empty_global_costmap_params.yaml" command="load" />
```
tangemicioglu: We should remove these config files as well since they were just temporary solutions until the changes were done.

igvc_navigation/config/teb_local_planner_params.yaml
oswinso: Yeah thats true. I'll revert the change.

igvc_navigation/config/global_costmap_params.yaml
oswinso: Yeah we should be able to

igvc_navigation/launch/refactored_mapper.launch
oswinso: Yupp

PR #539, opened 2019-10-13
Fixes new nodes not publishing and TEB fail scenario

### LINKED ISSUE #532: Teb doesn't publish anything

tangemicioglu:
# Description

Teb's subscribers broke, and there were scenarios when it ended up with infeasible trajectories.

This PR does the following:
- Fixes new nodes' subscribers
- Makes the global planner actually plan
- Removes Field D* from TEB launch file

Fixes #532 

# Testing steps (If relevant)
## Test Case 1
1. Launch `qualification.launch` and `navigation_simulation_teb.launch`

Expectation: TEB should be able to navigate to waypoints without any issues (and without going over lines!).

# Self Checklist


igvc_navigation/src/action_server/action_server.h
```diff
@@ -19,6 +19,8 @@ class ActionServer
 {
 public:
   ActionServer();
+  ros::Subscriber rviz_sub;
+  ros::Subscriber waypoint_sub;
```
oswinso: Actually according to ROS style conventions members should end with a `_`, so ie. `rviz_sub_` and `waypoint_sub_`

PR #568, opened 2019-10-31
Add competition map and refactor simulation

### LINKED ISSUE #526: Create actual course in gazebo
Right now in gazebo the only "course" that we have is the qualifications course, launched with `roslaunch igvc_gazebo qualification.launch`. However, this is different from the real course.

AC: Create a gazebo world that models the real course shape, potentially with a ramp.

tangemicioglu:
# Description

The qualification map is too easy now, so this adds an autonav map roughly copied from this year's competition as well as an even more difficult challenge track. It also refactors the simulation launchfiles to make it easier to add new worlds.

This PR does the following:
- Adds a complete new competition map
- Refactors simulation launchfiles

Fixes #526 

# Testing steps (If relevant)
## Test Case 1
1. Make sure we can still run old map after refactor: `roslaunch igvc_gazebo qualification.launch`
2. The default for sim should also be qualification on track 0: `roslaunch igvc_gazebo simulation.launch`
3. Finally, try launching the new map: `roslaunch igvc_gazebo autonav.launch`

Expectation: All of the maps launch properly, with waypoints and accepting proper parameters

# Self Checklist


igvc_description/urdf/worlds/autonav.world
```diff
@@ -0,0 +1,554 @@
...
+    <scene>
+      <sky>
+        <clouds>
+          <speed>12</speed>
+        </clouds>
+      </sky>
+    <ambient>1.0 1.0 1.0 1.0</ambient>
+    <shadows>true</shadows>
+    </scene>
+
```
oswinso: You should specify the default camera coordinates so it points at the Jessii instead of at the origin.
http://sdformat.org/spec?ver=1.6&elem=world#gui_camera

oswinso:
One issue I'm having is that I've got a pretty low realtime factor. I'll make an issue for this, but we should have a parameter that allows us to run simulation faster at the cost of accuracy / update rate of sensors etc.

tangemicioglu:
The map is actually smaller than the qual map, so it's probably the amount of barrels that's causing the speed issue

oswinso:
Yeah. Don't worry about the performance for this PR, we'll handle it in a separate PR. Just get the default camera change in and this PR should be good.

Issue #130, opened 2016-10-24
Joystick tests currently failing (the only tests we have)


rmkeezer:
They have been failing for a while now, but circleci was not recognizing failed tests until now.

jgkamat:
This can probably be closed @rmkeezer

rmkeezer:
Good thinking

PR #573, opened 2019-11-01
IOP Research Doc

### LINKED ISSUE #554: Try out JausToolset and write a research doc for deciding whether to IOP
We need to decide whether or not we want to do IOP or not before the end of the semester, which relies on whether or not it is worthwhile.

AC: Try out JausToolset and write a research doc detailing what the experience is like using the library, and thoughts on whether we should do IOP or not.

oswinso:
# Description
So, to my knowledge we've kind of just ignored the IOP part of the challenge since like forever (I searched through all git history for the words IOP and nothing showed up. Maybe we used to before moving to github, but like that's way too long ago and doesn't count). We decided to evaluate whether or not it would be worth doing IOP, and so here is a research doc that summarizes what I have found about my IOP, and my thoughts on whether I think we should do it or not.

[Link to doc](https://github.com/RoboJackets/igvc-software/blob/research/iop/documents/research/iop.md) (Please don't try and read a markdown file from the diff)

This PR does the following:
- Adds a research doc for IOP to the repository 

Fixes #554 

# Self Checklist

### DIFF
--- a/documents/README.md
+++ b/documents/README.md
@@ -3,6 +3,7 @@ This folder contains research and design documents for the repo.
 
 ## Research Docs
 * [Barrel detection](research/barrel_detection.md)
+* [IOP](research/iop.md)
 
 ## Design Docs
 * [Mapping refactor](design/mapping.md)
--- a/documents/research/README.md
+++ b/documents/research/README.md
@@ -1,2 +1,3 @@
 # Research Docs
 * [Barrel detection](barrel_detection.md)
+* [IOP](iop.md)
--- a/documents/research/iop.md
+++ b/documents/research/iop.md
@@ -148,6 +148,17 @@ So we need two JAUS components. The services that are required of each component
 
 All of which are standard services defined in JAUS.
 
+There are 2 main parts to the scoring:
+1. IOP Interfaces Task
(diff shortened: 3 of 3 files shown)


oswinso:
Oh no #130 (TIL this was is a 3 year old issue) comes to bite us back.

JasonGibson274:
We should not buy an software for this. That's not worth the money at all.

Other notes, as much as I am entertained by the tenor of this document, what level of professionalism should be in these documents? I am generally okay with whatever I just want a consistent standard. This reads a lot like a stream of consciousness is mostly unclear to me at least. I cannot separate out what is your opinion and the actual solution is not clear to me at all. This is not skimmable at all and I think that is a problem, when implementing a system like design documents.

oswinso:
Yeah I totally agree. I was pretty frustrated and tired when I wrote this, so a lot of my opinions came out, but I'll try and remove a lot of the unnecessary things.

I am very confused by how you are confused by the actual solution though. What do you mean by that?

I think some opinions in the research documents is fine (There needs to be some interpretation / summary / synthesis of the information obtained, otherwise I don't think there is much point in writing these documents if you don't come out having had a more well informed opinion).

PR #581, opened 2019-11-04
Fix bug with costmap converter not producing output + Initial tuning

### LINKED ISSUE #580: Costmap converter not working
The costmap converter seems to not be producing any output.

Steps to reproduce:
1. Launch simulation
2. `roslaunch igvc_navigation navigation_simulation.launch`
3. Launch rviz
4. Turn on `/move_base_flex/TebLocalPlannerROS/teb_markers`

Expected behaviour: `teb_markers` should show a bunch of polygons representing the obstacles
Actual behaviour: `teb_markers` doesn't show anything

The reason for this seems to be that the local costmap large, but the costmap converter has no option for limiting how many obstacles it tries to convert, so it ends up converting everything, which is quite computationally expensive??

oswinso:
# Description
After some changes (I haven't bisected so I don't actually know which change it was that introduced the bug), the costmap converter doesn't seem to be working anytmore.

This PR does the following:
- Allows the `lidar_layer` and `line_layer` to work with the `rolling_window` settings, so that we can have a smaller local costmap
- Creates a `UnrollingLayer` that takes in the output of the local costmap (which is rolling), and uses that to build back the large costmap again

Fixes #580 

EDIT: I've also accidentally started tuning TEB while waiting for the PR, so sorry for the larger commit.

# Testing steps (If relevant)
## `teb_markers` shows polygons representing obstacles
2. Add some barrels to the front (ugh why does the default qualification.launch not have barrels in the default lane
3. Launch the navigation stack. `roslaunch igvc_navigation navigation_simulation.launch`
4. Open up rviz.
5. Add the `/move_base_flex/TebLocalPlannerROS/teb_markers` topic to rviz

Expectation: rviz should show pink polygons represented to the obstacles

# Self Checklist


tangemicioglu:
I'm getting a delay of several seconds for updates to the global costmap. Any ideas?

oswinso:
I'm not sure I understand what you mean.

Maybe it's the changed behavior for the global map? So, because our local map is now a rolling window, our global map basically only updates using the rolling window.

tangemicioglu:
I realized this when overlaying the markers with the global costmap, but there's always a delay that takes a very visible amount of time for the global costmap to update after the local costmap updates. This reduces the path planner's response speed to newly detected obstacles. Maybe there's a low frequency to the way we're updating with this?

oswinso:
How much is "very visible"? On my desktop, it seems to be pretty instant, ie. short enough that it shouldn't really affect that much.

oswinso:
@tangemicioglu any updates?

tangemicioglu:
Rviz is giving "Uninitialized quaternion" warnings for all the detected PolyObstacles. The delayed updates in global costmap issue seems to be worse, I had no updates for 20 seconds at some point in qualification.

oswinso:
Wait whaaat? I'll try and check up with you today after the meeting to see whats up.

igvc_navigation/src/mapper/unrolling_layer.cpp
```diff
@@ -0,0 +1,249 @@
...
+  min_map_x_ = std::min(min_map_x_, start_x);
+  min_map_y_ = std::min(min_map_y_, start_y);
+  max_map_x_ = std::max(max_map_x_, end_x);
+  max_map_y_ = std::max(max_map_y_, end_y);
+}
+
+void UnrollingLayer::updateBounds(double robot_x, double robot_y, double robot_yaw, double* min_x, double* min_y,
+                                  double* max_x, double* max_y)
+{
+  ROS_INFO_STREAM(">>> updateBounds");
```
tangemicioglu: Could you remove these info prints? It doesn't provide much information now that you're done debugging and they get very spammy in the info stream.

igvc_navigation/config/teb_local_planner_params.yaml
```diff
@@ -33,7 +33,7 @@ TebLocalPlannerROS:
   # GoalTolerance
 
   xy_goal_tolerance: 0.2
-  yaw_goal_tolerance: 1.57
+  yaw_goal_tolerance: 3.14
```
tangemicioglu: As I mentioned in Software Core, this probably doesn't matter. The tolerance is a plus or minus thing so both of these should have the same effect. I'm also still getting curved paths when approaching waypoints in simulation.

tangemicioglu: Maybe we should open an issue to understand what's actually causing this? Doesn't really feel like a priority to me and could be added to the TEB tuning task instead of being an independent issue, but still good to make a note for later.

igvc_navigation/src/mapper/unrolling_layer.cpp
oswinso: oh woops I forgot that I added these in, mb

igvc_navigation/config/teb_local_planner_params.yaml
oswinso: yeah, that is true. Ok, I'll create a bug for it.

PR #583, opened 2019-11-06
Motor output from MBED

### LINKED ISSUE #403: Return motor controller PWM values

oswinso:
# Description

We currently have no idea what values our MBED is sending to the motors, which is kind of scary, and hard to debug controls problems.

This PR does the following:
- Creates a `igvc_msgs::motor_control` message type for modelling the MBED output to the motors (0 - 255)
- Updates the `igvc.pb.c` and `igvc.pb.h` to include the new `left_output` and `right_output` fields
- Updated `motor_controller` to extract the motor outputs and publish them.

Fixes #403.

# Testing steps (If relevant)
## Test on Jessii
1. Launch `motor_controller` on Jessii and a joystick node, with wheels decoupled
2. `rostopic echo /motor_output`
3. Move the joystick around

Expectation: It should values corresponding to what you are joysticking

# Self Checklist


igvc_msgs/msg/motor_control.msg
```diff
@@ -0,0 +1,4 @@
+Header header
+uint8 left
+uint8 right
+duration duration
```
Dallas-D: Is the first `duration` supposed to be capitalized?

igvc_platform/src/motor_controller/motor_controller.cpp
```diff
@@ -277,6 +279,14 @@ void MotorController::publishResponse(const ResponseMessage& response)
   enc_msg.duration = response.dt_sec;
   enc_msg.header.stamp = ros::Time::now() - ros::Duration(response.dt_sec);
   enc_pub_.publish(enc_msg);
+
+  // Publish motor output
+  igvc_msgs::motor_control motor_control;
+  motor_control.left = static_cast<uint8_t>(response.left_output);
+  motor_control.right = static_cast<uint8_t>(response.right_output);
+  motor_control.duration = ros::Duration(response.dt_sec);
+  motor_control.header.stamp = ros::Time::now() - motor_control.duration;
```
Dallas-D: I'm not sure why you subtracted duration from the current time. What do you want the header time to represent?

oswinso: You're right this should just be `ros::Time::now()`.

igvc_msgs/msg/motor_control.msg
oswinso: Nope. Check [here](http://wiki.ros.org/msg#Fields)

PR #591, opened 2019-11-09
Add max distance for line_layer

### LINKED ISSUE #590: Cap camera projection range
We need a maximum distance for our projections, [since anything longer than like 10m is pretty bad](https://www.youtube.com/watch?v=R6zcGU43RVs&feature=youtu.be).

AC: parameter for maximum distance of camera projections for mapping.

oswinso:
# Description
Our projections and our line segmentation are bad at longer distances, resulting in us "forgetting lines" at longer distance due to free space clearing.

This PR does the following:
- Adds a maximum distance to the line_layer
- Also sets "track_unknown" to false (but I don't want to make a new PR for this)

Fixes #590 

# Testing steps (If relevant)
## Maximum distance works
1. First, set `miss_exponential_coeff`, `miss_angle_exponential_coeff`, and `hit_exponential_coeff` to `0`, so that we don't get exponential decay due to distance for probabilities
2. Open up simulation
3. Open up rviz
4. Drive further than 10 meters from a line (Check rviz)
5. Open up our navigation stack. `roslaunch igvc_navigation navigation_simulation.launch`
6. Slowly drive forwards using a joystick or `rostopic pub`

Expectation: All lines that are within 10m are plotted, anything outside is ignored

# Self Checklist

### DIFF
--- a/igvc_navigation/config/global_costmap_params.yaml
+++ b/igvc_navigation/config/global_costmap_params.yaml
@@ -10,7 +10,7 @@ global_costmap:
     height: 200
     origin_x: -100
     origin_y: -100
-    track_unknown_space: true
+    track_unknown_space: false
     unknown_cost_value: 30
     rolling_window: false
     resolution: 0.1
--- a/igvc_navigation/config/local_costmap_params.yaml
+++ b/igvc_navigation/config/local_costmap_params.yaml
@@ -10,7 +10,7 @@ local_costmap:
     height: 200
     origin_x: -100
     origin_y: -100
-    track_unknown_space: true
+    track_unknown_space: false
     unknown_cost_value: 30
     static_map: false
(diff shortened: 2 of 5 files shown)


oswinso:
The reason why we "forget" lines is because:
1. We see a line somewhere
2. We go a far distance
3: We look back at the line, but because the line is so far away, it appears too small on the image to be picked up by the camera, and thus image segmentation (it should be theoretically visible though)
4: Because we "should" be seeing it, but our image segmentation doesn't pick it up because its too small in the image, we mark that area as free space
5: Because we mark it as free space, we decrease the probability that there is a line there, thus causing the line to disappear

By limiting the distance, at (5) even though it is free space, because the distance is so far away we don't decrease the probability, thus hopefully solving the issue.
@Dallas-D

PR #594, opened 2019-11-10
Add software lead icon to readme

Dallas-D:
# Description

Adds a fancy "Software Lead" icon to our readme (like [RoboRacing](https://github.com/RoboJackets/roboracing-software)). Looks like [this](https://github.com/RoboJackets/igvc-software/tree/feat/fancy_readme). We do not have a maintainer or a PM on GitHub, so those are excluded. Ours also has RoboBuzz, making it way better than RR.


oswinso:
I strongly prefer if you could make the PR and commits be more descriptive, ie. instead of
> Update readme.md
maybe something like
> Add software lead badge to readme.md
or something. Also, I'd prefer if you could create an issue before you make the PR, even if the issue doesn't exist previously, just so that we have a centralized place where are able to search for problems and have previous issues that may have addressed those problems show up

Dallas-D:
I don't think it's worth creating a new issue for this PR since it's relatively simple and unimportant. I went ahead and updated the PR so it can be referenced later.

oswinso:
Ok, but I still think that you should at least try to add a description to the PR

PR #601, opened 2019-11-21
Design/add noise to simulation

### LINKED ISSUE #600: Design Document for Add Noise to Simulation Issue
AC: Write a design document that solves the issue #552

w-illiam-wilson:
Contains the design document solving the problem of adding noise to the simulation

Fixes #600

### DIFF
--- a/documents/design/AddNoiseToSimulation.md
+++ b/documents/design/AddNoiseToSimulation.md
@@ -15,7 +15,8 @@ added to the ground truth pose, and this noise can be set in the gazebo launch f
 
 ## Proposed Solution
 
-I am going to simulate noise via random Gaussian noises (random normal distribution)
+I am going to simulate noise for the x and y position of the robot via random Gaussian noises
+(random normal distribution) in the ground_truth node.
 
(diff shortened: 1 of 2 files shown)


documents/design/AddNoiseToSimulation.md
```diff
@@ -0,0 +1,48 @@
...
+
+**Author:**
+- William Wilson
+
+## The Problem
+
+The problem I am trying to solve is the lack of noise in the gazebo simulation. In order to
+have a more realistic simulation and ground truth pose, noise is needed. In the real world,
+sensors output noisy data, which we must work through to get an accurate position of the
+robot. Currently, the simulation does not account for that. In the end, there will be noise
```
oswinso: Currently, the simulation actually does simulate noise. The problem is that our localization and mapping nodes are able to work with the current level of noise. The goal here is to create more of a "stress" test, ie.

If the localization node does a poor job and outputs a noisy / bad estimate, how does that affect the mapping stack? How does that affect the navigation stack?

Issue #571, opened 2019-10-31
Turn the cone of shame into the cone of fame


oswinso:
To prevent going backwards, we need to legitimize the [cone of shame](https://github.com/RoboJackets/igvc-software/tree/master/igvc_navigation/src/back_circle), and properly integrate it into the mapper.

We can do some OpenCV stuff to properly make a line that goes from the left to the right, instead of guessing how large the cone should be.

Ask @JasonGibson274 for details on the cone of shame.

![cone_of_shame](https://user-images.githubusercontent.com/4125463/67965541-97cfda80-fbd8-11e9-94de-4dbec2723d46.png)

AC: Jessii doesn't go backwards on the autonav simulation course.

tangemicioglu:
Based on behavior I've seen in autonav, I think we'll need the cone of fame at every waypoint.

JasonGibson274:
Really? That seems wrong, it should only really be required for the first way point I think. Can you attach a picture with waypoints, course, and proposed global path?

oswinso:
Yeah, I agree with Jason. For the second to third, the straight line should be the correct global bath. For the third to the fourth, because we already have the global map it's going to plan correctly.

tangemicioglu:
Second to third is the problematic one, the only "straight line" you get there is when the robot goes outside of the lines. I can't pull up the sim right now, but here's a drawing of what happens:
![ApplicationFrameHost_ioygBwZC3y](https://user-images.githubusercontent.com/11898470/68147434-04fcac00-ff08-11e9-8b5f-0656ca6d438d.png)

oswinso:
oh, interesting. I guess that makes sense. Yeah, we'll probably need to do something to make sure that that's not valid.

JasonGibson274:
Do we have to go across the bridge? That looks like a valid path?

oswinso:
I think we need to go through the center thing. I haven't checked the rulebook for a specific rule, but I get the feeling that we aren't supposed to go that way.

JasonGibson274:
I don't see that we have to over the ramp in the rules. Yes it is faster, but it is not required. You just need to get to the waypoints.

tangemicioglu:
That feels a bit cheaty especially now that they've removed the ramp and replaced it with a barrel course, but I can't find any rules forbidding us now that I look around the rulebook. Is there a way for us to ask the organizers about this scenario?

JasonGibson274:
[...]

PR #603, opened 2019-11-21
Cone of Shame Design Document

jiajunmao:
# Description

Design Document for fixing issue #571 

[Link to document](https://github.com/RoboJackets/igvc-software/blob/design/cone_of_shame/documents/design/cone_of_shame.md)

This PR does the following:
- Create the design document for fixing cone of shame issue

### DIFF
--- a/documents/design/cone_of_shame.md
+++ b/documents/design/cone_of_shame.md
@@ -13,9 +13,9 @@ In last year's IGVC competition, Jessie turned 180 degrees before starting trave
 ## Proposed Solution
 
 - The problem could potentially be solved by creating a cone behind the robot and make the robot think that that is a part of the line defining the course. Thus the robot would not travel backwards.
-- Create a new topic publishing a map with artificially created points and then feed that into the navigation 
-    - Create a map with artificially inserted points and publish the map to a topic (maybe something like /cam/fakeback)
-    - Create a subscriber in the navigation stack for the topic and interegrate the mapping into path planning cost map
+- Create a new layer in the 2D cost map and artificially insert points to act as a cone-shaped obstacle behind the robot
+    - Create a new layer and artificially create point behind the robot
+    - Incorporate the new layer into the cost map
 
 ## Questions & Research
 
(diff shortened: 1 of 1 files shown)


documents/design/cone_of_shame.md
```diff
@@ -0,0 +1,39 @@
...
+
+## The Problem
+
+In last year's IGVC competition, Jessie turned 180 degrees before starting traveling on the course. This is presumably caused by the current navigation model ignoring everything behind Jessie and default that nothing exists there. This caused the problem with the path planning algorithm thinking that the shortest path is to walk backward to the place where "no line exists".
+
+## Proposed Solution
+
+- The problem could potentially be solved by creating a cone behind the robot and make the robot think that that is a part of the line defining the course. Thus the robot would not travel backwards.
+- Create a new layer in the 2D cost map and artificially insert points to act as a cone-shaped obstacle behind the robot
+    - Create a new layer and artificially create point behind the robot
```
oswinso: Give a bit more detail about how this is going to work, ie.

When are we going to create points behind the robot? Are we going to remove them?
What shape are these points going to be? How large?

documents/design/cone_of_shame.md
```diff
@@ -0,0 +1,39 @@
...
+    - Incorporate the new layer into the cost map
+
+## Questions & Research
+
+- I probably just need to go through the input of navigation stack to make sure the specific format I am supposed to be publishing and the format for the name of the topic
+
```
oswinso: Should also add in [how to create a costmap2d layer](http://wiki.ros.org/costmap_2d/Tutorials/Creating%20a%20New%20Layer) and [how ROS services work](http://wiki.ros.org/ROS/Tutorials/WritingServiceClient%28c%2B%2B%29)

documents/design/cone_of_shame.md
jiajunmao: Details have been updated

documents/design/cone_of_shame.md
jiajunmao: I have added in details on how costmap2d layer and ROS service work

documents/design/cone_of_shame.md
```diff
@@ -0,0 +1,55 @@
...
+## The Problem
+
+In last year's IGVC competition, Jessie turned 180 degrees before starting traveling on the course. This is presumably caused by the current navigation model ignoring everything behind Jessie and default that nothing exists there. This caused the problem with the path planning algorithm thinking that the shortest path is to walk backward to the place where "no line exists".
+
+## Proposed Solution
+
+- The problem could potentially be solved by creating a cone behind the robot and make the robot think that that is a part of the line defining the course. Thus the robot would not travel backwards.
+- How to measure the location of the cone
+    - Subscribe to **/move_base_flex/local_costmap/costmap** and select the point on the line that is closest to the robot
```
oswinso: Since this is going to be in a `costmap_2d::Layer`, you don't actually need to subscribe to `/move_base_flex/local_costmap/costmap` to get access to the `line_layer` (it's a bit roundabout). The `costmap_2d::layer` base class has a `layered_costmap_` member, which in turn contains a `getPlugins` member function returning `std::vector<boost::shared_ptr<Layer> >*`. You should be able to check if the the `boost::shared_ptr<Layer>` is of type `LineLayer` by using `boost::dynamic_ptr_cast`, after which you can just work with the `LineLayer`'s internal map representation directly. The benefit of this is that you don't need to serialize + deserialize the map (expensive), and that there will only be lines now as opposed to lines and barrels.

jiajunmao:
I have addressed the two additional details concerned.

PR #607, opened 2019-11-21
Feat/barrel detection open cv

PCApple:
# Description

This PR should be able to identify barrels using openCV and remove them from the freespace pointcloud. It kind of works along linelayer.cpp

This PR does the following:
it is going to first create 2 buffers up at the top, barrel_buffer_ and _not_barrels_. Then, in the imageSyncedCallback, it will create another Mat called barrel_mat using a function called findBarrel. You can change the boolean to change if the image will be published to pauls_stuff/barrel

The findBarrel function is pretty simple, it will use a gaussian blur and then a HSV threshold to create the frame. It will open and close to cleanup, and then check if you want to debug, calling a function to send the frame out if the boolean is true. 

The project images function was edited slightly to use the barrel mat. it will edit the barrel_buffer as well to be used later in  the cleanup function. The barrels "true" value is 100U rather than 255U to make sure that they aren't confused. Each point will be checked if the are is a barrel, and if it is, it will mark it in the barrel_buffer_.

Finally, in the cleanupProjections(), I added some code to use bitwiseNot to create a not_barrels_ mat and to remove it from the freespace buffer.

I'd say to test it using rviz and the local costmap buffer to see where the lines are getting drawn. Also you can set the boolean in the findBarrel() call to true on line 135 to push the frames to pauls_stuff/barrels. This is a bit buggy, since it will send all 3 camera angles sequentially, so keep that in mind
Fixes #{{ 536}}

Expectation: The thing which does the thing does the thing


documents/design/openCV_barrel_detection.md
```diff
@@ -0,0 +1,67 @@
...
+
+Are there things you are unsure about or don't know? What do you need to research to be able to
+complete this project? If you need information from the mechanical or electrical subteams,
+be sure to describe that here. If your solution requires more research to implement, describe
+what kind of topics you need look at. Link any research papers/articles that you find here.
+
+**Of course, if you have _any_ questions or concerns, feel free to bring them up at any time.
+If you ever feel lost or don't know what to do or work on, Oswin or any old software member
+can help you out. You are not expected to know every technical detail about the project. Expressing
+what you don't know will make it easier to do research and ask for help.**
```
oswinso: Delete this

PR #611, opened 2019-11-24
CNN.

### LINKED ISSUE #504: cnn.py doesn't conform to PEP8
I'm getting some errors about `cnn.py` not conforming to PEP8 standards, ie.

```python
transform = transforms.Compose([
                 transforms.ToTensor(),
            ])

class SegmentationModel(object): # PEP8: expected 2 blank lines, found 1
```

Let's make `cnn.py` conform to PEP8.

AC: No PEP8 errors in `cnn.py`.

suhacker1:
# Description

These changes make cnn.py conform to PEP 8 using Flake8 as linter. It might be useful if someone else also runs cnn.py to make sure everything works. 

This PR does the following:
- Fixes line lengths, spacing, import orders, and other issues linter pointed out 

Fixes #504 

# Testing steps (If relevant)
## Test Case 1
1. Run cnn.py 
2. Apply Flake8/linter

Expectation: Flake8/Linter shows no errors regarding PEP 8 conformity 

# Self Checklist
- [N/A] I have formatted my code using `make format`
- [X ] I have tested that the new behaviour works


igvc_perception/src/cnn/cnn.py
```diff
@@ -1,42 +1,40 @@
-#!/usr/bin/env python
```
oswinso: We need the shebang for ROS. Also, I don't think this violates any PEP8 stuff.

igvc_perception/src/cnn/cnn.py
```diff
@@ -1,42 +1,40 @@
...
+from cv_bridge import CvBridge
 from image_geometry import PinholeCameraModel
 from sensor_msgs.msg import CameraInfo
 from sensor_msgs.msg import CompressedImage
 from sensor_msgs.msg import Image as ImMsg
 from sensor_msgs.msg import PointCloud2
 from sensor_msgs.msg import PointField
 
+sys.path.append(os.path.dirname(os.path.abspath(__file__)) + "/../")
+
```
oswinso: We actually can't have this below the `from train_eval.models.model import UNet`, as if you look at the directory structure, `train_eval` is like located in the parent directory, if you know what I mean.

Anyways, I think the proper way of fixing this is to remove this line, and then move the `train_eval` directory inside the `cnn` directory, so we can do `from train_eval.models.model import UNet` normally without any jank path manipulations

oswinso: Also, it turns out that the rest of our files are have a bunch of PEP8 violations... 😅

suhacker1:
For some reason, PEP8 (at least pep8speaks) doesn't like the shebang, but perhaps we should just ignore that convention for the sake of functionality.

PR #615, opened 2019-11-25
Add a 3D mesh map for Autonav

### LINKED ISSUE #544: Heightmap for gazebo
We should simulate a nonflat terrain by adding in a heightmap for gazebo.

[This](https://github.com/AS4SR/general_info/wiki/Creating-Heightmaps-for-Gazebo) looks like a easy guide to do this.

AC: Non flat ground in gazebo

tangemicioglu:
# Description

Since the heightmap breaks Velodyne, the alternative (and apparently better, based on the processing time) option is to make a 3D mesh for a more realistic ground surface. This adds a map designed in Blender (and available in the Robojackets Cloud) and a new launchfile for it. This PR also gets rid of some deprecated components in the simulation and description to clean things up.

This PR does the following:
- Creates a new 3D world map with a mesh
- Gets rid of references to jaymii, real_igvc and terrain

Fixes #544 

# Testing steps (If relevant)
## Test Case 1
1. Launch sim: `roslaunch igvc_gazebo autonav_mesh.launch`
2. Launch navigation: `roslaunch igvc_navigation navigation_simulation.launch`

Expectation: The new map works, displays textures and the robot can navigate in it.

# Self Checklist


tangemicioglu:
I lowered the plane a little, and now all the barrels should safely fall into place instead of clipping through and bouncing to the middle of the track.

oswinso:
The ramp _still_ seems to be floating.... :(

Screenshot from 2020-01-15 00-08-24

Try double checking by driving over the ramp using a joystick. If we can drive over the ramp with a joystick then it should be fine.

PR #620, opened 2019-12-02
Add noise to simulation

### LINKED ISSUE #552: Add noise to simulation
We need to have noisier data to make the simulation more realistic. We should add noise to the ground truth pose of the robot and make it configurable so that we can change the amount of noise as we wish.
AC: We can set the amount of noise for ground truth in the igvc_gazebo launchfiles.

w-illiam-wilson:
# Description

This adds a randomizer function to ground_truth with the purpose of adding noise directly to ground_truth data. The randomizer's standard deviation (noise) can be set in simulation.launch.

This PR does the following:
- Adds a randomizer function to ground_truth and adds it to the x and y positions
- Adds a launch parameter for the standard deviation of said randomizer to simulation.launch

Fixes #552 

# Testing steps (If relevant)
## Test Case 1
1. Play around with noise parameter in simulation.launch
2. rostopic echo/ground_truth and check position variables

Expectation: Noise is added directly to ground_truth positions

# Self Checklist


igvc_gazebo/launch/simulation.launch
```diff
@@ -18,6 +18,7 @@
 
     <param name="/use_sim_time" value="true"/>
     <param name="/waypoint_file_path" value="$(arg waypoints)" />
+    <param name="noise" type="double" value="0.0"/>
```
oswinso: I think you should make this parameter a bit more descriptive. Does this represent the standard deviation / variance of the noise? On what values (x, y, z, heading) does it get applied to?

igvc_gazebo/nodes/ground_truth/main.cpp
```diff
@@ -1,32 +1,38 @@
+#include <functional>
 #include <geometry_msgs/Vector3.h>
+#include <iterator>
+#include <mutex>
```
oswinso: I don't think the `include <mutex>` here is necessary as I don't think you are using any mutexes here. I know you didn't add this, but I think since you're changing up this file it's a good idea to clean it up.

igvc_gazebo/nodes/ground_truth/main.cpp
```diff
@@ -81,29 +88,32 @@ void utm_callback(const ros::TimerEvent& event, const tf::Transform& odom_to_utm
...
-int main(int argc, char** argv)
+int main(int argc, char **argv)
 {
   ros::init(argc, argv, "ground_truth_republisher");
 
   ros::NodeHandle nh;
   ros::NodeHandle pNh("~");
 
+  nh.getParam("/noise", noise);
+
```
oswinso: You should use the `assertions::parm` helper function to make sure that the node errors out if this is not set.

igvc_gazebo/nodes/ground_truth/main.cpp
```diff
@@ -1,32 +1,38 @@
...
+void odomCallback(const nav_msgs::Odometry::ConstPtr &msg)
 {
   g_last_estimate = msg->header.stamp;
 }
 
-void groundTruthCallback(const nav_msgs::Odometry::ConstPtr& msg)
+void groundTruthCallback(const nav_msgs::Odometry::ConstPtr &msg)
 {
+  // random noise generator
+  auto dist = std::bind(std::normal_distribution<double>{ 0.0, noise }, std::mt19937(std::random_device{}()));
```
oswinso: I think this is a bit inefficient, as you're creating a new `std::random_device` every time, when you really only need it to seed the `std::mt19937` PRNG. As a whole, I think you could clean up this node by making it a class, but it's fine if you leave it as it is, it's just that with globals it's a bit more messy.
I would suggest that you put the `std::mt19937` as a global / member variable, so that you only create the `std::random_device` once, and then keep reusing the same `std::mt19937`.

oswinso: Another thing is that I think it would be nice to have different noise variances for the different fields, ie. x, y, heading, so we can see which fields our entire system are most sensitive to noise to.

igvc_gazebo/nodes/ground_truth/main.cpp
```diff
@@ -1,32 +1,38 @@
...
   // get the starting location as the origin
   if (g_og_pose.header.stamp.toSec() == 0)
   {
     g_og_pose.pose = msg->pose;
```
oswinso: With the comment on having different noise variances for different fields, you can add noise to the heading by accessing the `orientation` member here. Remember to use the `tf::getYaw` and `tf::createQuaternionMsgFromYaw` helpers so that you can convert to / from quaternions.

oswinso:
Closing as its abandoned

Issue #622, opened 2019-12-08
No way to set `/waypoint_file_path` outside of simulation


oswinso:
With #613, we use `/waypoint_file_path` to set the path for the `.csv` waypoints file. However, there isn't a way to set this parameter for non-simulation right now.

AC: A way to set `/waypoint_file_path` to `.csv` in the sandbox folder easily.

PR #623, opened 2019-12-09
Fix Navigation Client Issues

### LINKED ISSUE #621: Navigation client doesn't parse comments nor whitespace properly
The navigation client doesn't parse comments (lines beginning with `#`) nor empty lines properly.

Steps to reproduce: Set `/waypoint_file_path` to `$(find igvc_navigation)/../sandbox/waypoints/parking_lot_loop.csv`.
Expectation: It should parse properly.
Bug: It errors out on `ROS_ERROR_STREAM(waypoint_file_path_ << ":" << line_index << " - " << tokens.size() << " tokens instead of 2.");`

AC: It can parse comments and empty lines.

matthewhannay567:
# Description

Fixes two issues with `navigation_client`

This PR does the following:
- Adds a new launch file to `igvc_navigation` called `set_waypoint_file_path.launch` which takes an arg `file_path` and sets the `waypoint_file_path` parameter to the value of `file_path`.
- Changes `simulation.launch` to use `set_waypoint_file_path.launch`.
- Fixes `NavigationClient::parseWaypointFile` not checking for comments or empty lines.

Fixes #621 and #622 

# Testing steps (If relevant)
## Test Case 1
1. run `roslaunch igvc_navigation set_waypoint_file_path.launch folder_path:=abc file_name:=def`
2. run `rosparam get /waypoint_file_path

Expectation: `abc/def` is displayed.

## Test Case 2
1. run `roslaunch igvc_navigation set_waypoint_file_path.launch file_name:=abc`
2. run `rosparam get /waypoint_file_path`

Expectation: `$PATH_TO_IGVC_SANDBOX/waypoints/abc` is displayed.

## Test Case 3
1. run `roslaunch igvc_navigation navigation_client.launch` while a simulation that uses a waypoint file with comments and clear lines.

Expectation: navigation_client does not break while loading waypoints

# Self Checklist
- [x] I have tested that the new behavior works


igvc_navigation/src/navigation_client/navigation_client.cpp
```diff
@@ -93,27 +93,44 @@ std::vector<LatLong> NavigationClient::parseWaypointFile()
...
 
   return coordinates;
 }
 
+std::string NavigationClient::removeWhitespace(const std::string& str)
+{
+  std::string new_string;
+  for (char c : str)
+  {
+    if (c != ' ')
```
oswinso: if you're doing this then I think you should just use
```
str.erase (std::remove (str.begin(), str.end(), ' '), str.end());
```

igvc_navigation/launch/set_waypoint_file_path.launch
```diff
@@ -0,0 +1,4 @@
+<launch>
+    <arg name="file_path"/>
+    <param name="/waypoint_file_path" value="$(arg file_path)"/>
+</launch>
```
oswinso: For ease of use, I think it would be nice to be able to split this into like a "folder" and a "filename", so that instead of having to specify the entire path on the command line, we can specify just the name of the `.csv`

oswinso:
I'm getting an error when I try to run simulation:
```
RLException: while processing /home/oswinso/catkin_ws/src/igvc-software/igvc_gazebo/launch/simulation.launch:
while processing /home/oswinso/catkin_ws/src/igvc-software/igvc_navigation/launch/set_waypoint_file_path.launch:
Invalid <arg> tag: sandbox
ROS path [0]=/opt/ros/melodic/share/ros
ROS path [1]=/home/oswinso/catkin_ws/src
ROS path [2]=/opt/ros/melodic/share. 

Arg xml is <arg default="$(find sandbox)/waypoints" name="folder_path"/>
The traceback for the exception was written to the log file
```

this is because you have `$(find sandbox)` in the `<arg name="folder_path" default="$(find sandbox)/waypoints"/>`, but `$(find X)` only works for packages, and rn `sanbox` isn't a package. I think the best solution here is to make `sandbox` a package (probably rename sandbox to `igvc_sandbox` or something, and then make a `package.xml` to get ROS to recognize the folder as a package.

oswinso:
Another thing is that you probably want a default arg for `reading_from_file` for `navigation_client.launch` so that you don't have to specify it everytime you want to run it (I think a sane default would be to have `reading_from_file` be `false`)

Issue #551, opened 2019-10-21
Add flake8 to CI


oswinso:
We've got clang-format for formatting our C++ side, but we've got nothing for python.

Check [this](https://realpython.com/python-continuous-integration/) out for introduction to flake8.

AC: flake8 added to CI

suhacker1:
I think we can consider this resolved since we added Pep8 Speaks?

oswinso:
Closed as #625 confirms that pep8speaks works.

PR #625, opened 2020-01-05
[IGNORE] Add very long line to cnn.py to test pep8speaks

oswinso:
Ignore this PR.

Trying to test if pep8speaks works.


oswinso:
Looks like pep8speaks is working properly, so we can close #551.

PR #626, opened 2020-01-08
Firmware Refactor Design Document

### LINKED ISSUE #588: Design Document on Firmware Rewrite
I'd like us to rewrite our firmware and communications code, on both the PC and the mbed side. Before we do that though, I think we should do a design document to properly design this out.

AC: Design document for firmware and communications between PC and MBED.

Dallas-D:
# Description

This PR is the design document for the firmware rewrite planned for spring 2020. The document itself can be read [here](https://github.com/RoboJackets/igvc-software/blob/design/firmware/documents/design/FirmwareRefactor.md).

There are the major changes proposed in the document:
- A new multi-threading scheme for improved efficiency
- Code structure changes for improved readability and maintainability
- New messages for communicating between the computer and the mbed
- More features and design problems to be aware of when implementing the system

Fixes #588

### DIFF
--- a/documents/design/FirmwareRefactor.md
+++ b/documents/design/FirmwareRefactor.md
@@ -1,0 +1,189 @@
+# Firmware Refactor Design Document
+
+Author: Dallas Downing
+
+## Motivation
+
+The firmware for IGVC has largely been neglected for the past few years, being an afterthought once more pressing problems have been addressed. However, we have reached a point where the basic code running on the mbed can no longer support the needs of the team. The goal of this firmware redesign is to tackle some of the main challenges with the firmware at last competition and make it easier to maintain and alter in the future. This will also (hopefully) motivate a new incentive to apply actual design decisions in making changes to the code later.
+
+Here is a short list of the problems that have been identified with the firmware and the mbed. Note that this is not a comprehensive list as many issues depend on the electrical system and the main computer. Most of the issues are explained later in the design section.
+
+All mbed code is contained in one `main.cpp` file, which makes understanding and editing the code difficult
+Asynchronous networking code (for communicating with the computer) runs in the same thread as “real-time” code (PID control)
+There are no errors messages sent from the mbed to the computer, making it difficult to see what the mbed is doing when it fails or encounters a problem
+Messages sent between the computer and the mbed follow the same format with many optional fields instead of using dedicated message formats for different commands
+The E-stop signal is treated as a normal digital input but should be an interrupt input to stop the robot as soon as possible
+
+There are also a number of new features being added to the electrical system that the mbed now must support. Part of the redesign was made keeping these features in mind. The specifics of their implementation has been left out of this document, but it should be clear how one might implement such features using the new firmware.
+A “virtual bumper” system using lidar lites that can warn and stop the robot preemptively before it hits an obstacle
+A diagnostics system with various sensors for detecting hardware issues with the robot 
+
+## Background
+
+The mbed is one of the most important components of the robot. Without it, the computer would just be a fancy computer and the rest of the robot would be a metal box filled with wires and silicon. The mbed is the bridge that connects the computer with the rest of the electrical system. This section will briefly go over what the mbed does to provide context for the upcoming design changes.
+
+Here are the main tasks that it performs:
+
+- Connects to the computer to receive PID values and motor commands (in the form of speeds for the left and right motor)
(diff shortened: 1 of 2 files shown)


documents/design/FirmwareRefactor.md
```diff
@@ -0,0 +1,189 @@
...
+
+The mbed is one of the most important components of the robot. Without it, the computer would just be a fancy computer and the rest of the robot would be a metal box filled with wires and silicon. The mbed is the bridge that connects the computer with the rest of the electrical system. This section will briefly go over what the mbed does to provide context for the upcoming design changes.
+
+Here are the main tasks that it performs:
+
+- Connects to the computer to receive PID values and motor commands (in the form of speeds for the left and right motor)
+- Performs PID calculations for bringing the motors to the right speeds
+- Receives signals from the encoders and keeps track of motor speeds and direction
+- Sends serial commands to the motor controller to properly drive the motors
+- Checks for when the e-stop is triggered and disables the motors when it is
```
JasonGibson274: It also clears constants from the PID (notably I, and D accumulation). This is actually more important than setting the speed to be 0 since the e stop system prevents the motors from running anyway.

Dallas-D: Yes, I left that part out of the document for clarity for non-software people. If we write more concrete documentation about the mbed, then this should definitely be put there.

Dallas-D: Yes, I left that part out of the document for clarity for non-software people. If we write more concrete documentation about the mbed, then this should definitely be put there.

documents/design/FirmwareRefactor.md
```diff
@@ -0,0 +1,190 @@
...
+
+Here is a short list of the problems that have been identified with the firmware and the mbed. Note that this is not a comprehensive list as many issues depend on the electrical system and the main computer. Most of the issues are explained later in the design section.
+
+- All mbed code is contained in one `main.cpp` file, which makes understanding and editing the code difficult
+- Asynchronous networking code (for communicating with the computer) runs in the same thread as “real-time” code (PID control)
+- There are no errors messages sent from the mbed to the computer, making it difficult to see what the mbed is doing when it fails or encounters a problem
+- Messages sent between the computer and the mbed follow the same format with many optional fields instead of using dedicated message formats for different commands
+- The E-stop signal is treated as a normal digital input but should be an interrupt input to stop the robot as soon as possible
```
oswinso: I think it should be clarified here that the mbed doesn't actually stop the robot as the estop electrically stops the motors.

documents/design/FirmwareRefactor.md
```diff
@@ -0,0 +1,190 @@
...
+A `MotorController` wrapper class will be created to handle communicating with the motor controller and will abstract its functionality. This change will accomplish 2 things: declutter the code for sending commands to the motor and achieve the overall software goal of writing self-documenting code. Here is how the motor controller is currently called in the code:
+
+``` Serial g_serial(p13, NC, 9600); ```
+
+This does nothing to explain what it is or how it works. A similar issue can be found in the PID control loop. The latter half of the `pid()` method is dedicated to converting the speed into motor values. This code should be taken out of that method and moved to the MotorController class. Here are the base functions that the `MotorController` class should provide:
+
+- Set the speeds of left and right motors independently
+- Get the speeds sent to the motors
+- Stop the motors
+- Advanced control of the motors (slowly stop the motors, turn around, etc.)
```
oswinso: Not too sure about this one, as I don't see anything that will make use of this

documents/design/FirmwareRefactor.md
```diff
@@ -0,0 +1,190 @@
...
+
+#### Status Lights
+
+The latest revision of the logic board includes pins for external status LEDs (besides the on-board LEDs). These and the existing LEDs were excluded from this document because the new functionality of the firmware can be defined before deciding what 
+
+---
+
+## Required Work
+
+The section will break down the implementation of this design document into _ major groups. This is intended to help create a timeline for the order of work and help assign people to different tasks.
```
oswinso: Fix the `_`

documents/design/FirmwareRefactor.md
Dallas-D: I put that there more as an idea for potential uses of the class beyond just basic functionality. I don't expect us to do anything like that unless we were going to do something very fancy with controls.

Issue #616, opened 2019-11-29
Design Document for Navigation State Machine


matthewhannay567:
AC: Design document for #520 / #596

PR #630, opened 2020-01-09
Feat/navigation state machine

### LINKED ISSUE #520: Create navigation state machine
With #518 we got a new node that translates `/waypoint` into actionlib requests to the `/move_base_flex/move_base`, which is a backwards compatible interface for `move_base`. [See more details about the move_base_flex actionlib interfaces here](https://raw.githubusercontent.com/magazino/move_base_flex/master/doc/images/move_base_flex.png) and [here](https://wiki.ros.org/move_base_flex#Concepts).

One problem with using the old move base interface (and why we chose to use `move_base_flex`) is that the global planner only plans a path *once* at the beginning of the request. Instead, we would like the global path to be constantly replanning, as our map is constantly changing.

We can do this by using the `/move_base_flex/get_path`, `/move_base_flex/exe_path` and `/move_base_flex/recovery` actionlib servers manually.

As for the state machine, I don't really have that great of an idea of how we should approach writing a good one. The `move_base_flex` docs have an [example of using SMACH to write a state machine](http://wiki.ros.org/move_base_flex/Tutorials/SimpleSmachForMoveBaseFlex), though this is in Python. I think we could also go with a hand-rolled one (our state machine isn't really that complicated) or use a library such as [Boost MSM](https://www.boost.org/doc/libs/1_71_0/libs/msm/doc/HTML/index.html) or [Boost Statechart](https://www.boost.org/doc/libs/1_71_0/libs/statechart/doc/index.html).

For now though, we just need a simple "state machine" that just calls the global planner and then passes that to the local planner (without any recovery behavior, that'll be another issue (TODO)).

AC: A "state machine" that for now just calls the global planner then passes that to the local planner at every timestep, mimicking our old behavior.

### LINKED ISSUE #596: Final orientation of waypoint for TEB is wrong
Currently, our waypoints disregard orientation and are fed directly into TEB. However, because TEB directly optimizes using the final orientation and only uses `yaw_goal_tolerance` for checking if it can finish the actionlib call, the resulting path does take the waypoint's orientation into account.

Thus, we need to either change the waypoint's orientation, or post-process the orientation of the goal pose for the global planner.

AC: Jessii no longer keeps turning towards 0 degrees on every waypoint.

matthewhannay567:
# Description

Implements a custom version of the move_base_flex navigation state machine.

This PR does the following:
- Adds the navigation_server object, which controls how the robot plans and moves.
- Fixes an issue where robot would approach waypoints from bad orientations.

Fixes #520 and Fixes #596 

# Testing steps (If relevant)
## Waypoint orientation correction
1. run `roslaunch igvc_gazebo qualification.launch`
2. move/rotate the robot to a new position (preferably closer to the center of the path)
3. run `roslaunch igvc_navigation navigation_simulation`

Expectation: The robot will approach the waypoint and not approach the waypoint from a strange angle.

# Self Checklist


oswinso:
Especially since you're implementing a state machine here, I think it would be really helpful to also have #616 done, so that we have a document (with graphics if possible) to detail the state machine transitions very clearly and explain how you designed this.

PR #638, opened 2020-01-19
fix for issue 637 cmath inclusion

### LINKED ISSUE #637: probability_utils.h not including cmath
probability_utils.h under igvc_navigation/include/mapper calls std::exp and std::log without including cmath and is causing build problems when running catkin_make.

jiajunmao:
# Description

This pull request is for fixing issue 637 where a non-inclusion of cmath in a file that calls std::log and std::exp is causing compile error when running catkin_make.

This PR does the following:
- just adding #include <cmath>

Fixes #637 

Expectation: catkin_make should pass without throwing error

# Self Checklist

### DIFF
--- a/igvc_navigation/include/mapper/probability_utils.h
+++ b/igvc_navigation/include/mapper/probability_utils.h
@@ -1,6 +1,8 @@
 #ifndef SRC_PROBABILITY_UTILS_H
 #define SRC_PROBABILITY_UTILS_H
 
+#include <cmath>
+
 namespace probability_utils
 {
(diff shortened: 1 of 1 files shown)


oswinso:
Remember the policy we're trying out this month:

> In the coming month, I'd like to try the following policy for PRs:
> **Assign someone other than me as the primary reviewer**. I will still make comments on the PR just to double check, but only after the assigned person has done their review and tested out the code first.

@jiajunmao assign someone else to double check

PR #639, opened 2020-01-24
Feat/recovery behavior

### LINKED ISSUE #597: Create recovery behavior
It turns out that [clear_costmap_recovery](http://wiki.ros.org/clear_costmap_recovery) works by reaching in to the layer's backing `costmap_2d` and clearing that, instead of doing it how it (IMO) should have been done by creating some "clearableLayer" thing that has a overridable method. So, we need our own recovery behavior. Also, we should probably think of what we should do exactly when global / local planning fails.

AC: Recovery behavior that works on our navigation stack.

### LINKED ISSUE #632: Implement Recovery Behavior
Our robot needs recovery behavior or its going to get stuck looking at a wall 50% of the time.
Which recovery behaviors should be considered / added?

matthewhannay567:
# Description

Implements a `back_up_recovery` recovery behavior, which makes the robot back up when it gets stuck.

This PR does the following:
- Adds `back_up_recovery` 
- Sets to it be the default recovery behavior for `navigation_server`

Fixes #597 and Fixes #632 

# Testing steps (If relevant)
## Test Case 1
1. Launch a simulation using `roslaunch igvc_gazebo autonav.launch`
2. Set the robot moving using `roslaunch igvc_naviation navigation_simulation.launch`

Expectation: The robot should attempt to turn around, get stuck on the wall, back up, and then turn around again and move forward.

# Self Checklist


igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
```diff
@@ -0,0 +1,100 @@
...
+  vel_msg.linear.x = -velocity_;
+  vel_pub.publish(vel_msg);
+
+  while (n.ok() && igvc::get_distance(current_position, start_position) < threshold_distance_ && next_step_valid)
+  {
+    local_costmap_->getRobotPose(global_pose);
+    current_position = global_pose.pose.position;
+    x = current_position.x - std::cos(current_yaw) * obstacle_distance_;
+    y = current_position.y - std::sin(current_yaw) * obstacle_distance_;
+    footprint_cost = world_model_->footprintCost(x, y, current_yaw, local_costmap_->getRobotFootprint(), 0.0, 0.0);
```
ceelo777: How does this footprint_cost work?

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+  geometry_msgs::PoseStamped global_pose;
+  local_costmap_->getRobotPose(global_pose);
+
+  double current_yaw = tf2::getYaw(global_pose.pose.orientation);
+  geometry_msgs::Point current_position = global_pose.pose.position;
+  geometry_msgs::Point start_position = current_position;
+  double x = current_position.x - std::cos(current_yaw) * obstacle_distance_;
+  double y = current_position.y - std::sin(current_yaw) * obstacle_distance_;
+  double footprint_cost = world_model_->footprintCost(x, y, current_yaw, local_costmap_->getRobotFootprint(), 0.0, 0.0);
```
ceelo777: Why is this repeated later on in lines 88-90? You could just initialize them with default values.

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
ceelo777: Why do you update the x and y positions for current_position before setting the velocity message?

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
```diff
@@ -0,0 +1,100 @@
...
+    y = current_position.y - std::sin(current_yaw) * obstacle_distance_;
+    footprint_cost = world_model_->footprintCost(x, y, current_yaw, local_costmap_->getRobotFootprint(), 0.0, 0.0);
+    next_step_valid = footprint_cost >= 0.0;
+
+    vel_pub.publish(vel_msg);
+    r.sleep();
+  }
+
+  vel_msg.linear.x = 0;
+  vel_pub.publish(vel_msg);
```
ceelo777: For each velocity message, why is there only a linear x-component and no angles or y + z?

igvc_navigation/src/back_up_recovery/back_up_recovery.h
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * @brief  Destructor for the back up recovery behavior
+   */
+  ~BackUpRecovery() override;
+
+private:
+  costmap_2d::Costmap2DROS* local_costmap_;
+  bool initialized_;
+  double velocity_, frequency_, threshold_distance_, obstacle_distance_;
+  base_local_planner::CostmapModel* world_model_;
```
ceelo777: Explain how Costmap Model works.

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
```diff
@@ -0,0 +1,100 @@
...
+    return;
+  }
+  ROS_WARN("Backup recovery behavior started.");
+
+  ros::Rate r(frequency_);
+  ros::NodeHandle n;
+  ros::Publisher vel_pub = n.advertise<geometry_msgs::Twist>("cmd_vel", 5);
+
+  geometry_msgs::PoseStamped global_pose;
+  local_costmap_->getRobotPose(global_pose);
```
ceelo777: How does getRobotPose() work?

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
```diff
@@ -0,0 +1,100 @@
...
+  vel_pub.publish(vel_msg);
+
+  while (n.ok() && igvc::get_distance(current_position, start_position) < threshold_distance_ && next_step_valid)
+  {
+    local_costmap_->getRobotPose(global_pose);
+    current_position = global_pose.pose.position;
+    x = current_position.x - std::cos(current_yaw) * obstacle_distance_;
+    y = current_position.y - std::sin(current_yaw) * obstacle_distance_;
+    footprint_cost = world_model_->footprintCost(x, y, current_yaw, local_costmap_->getRobotFootprint(), 0.0, 0.0);
+    next_step_valid = footprint_cost >= 0.0;
```
ceelo777: Why would the next step be valid if the footprint cost is greater than or equal to 0.0?

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
matthewhannay567: footprint_cost checks to see if there would be a cost (read: collision) if the robot were in that position

matthewhannay567: Velocity is constant, so the distance travel shouldn't change. The velocity is just set constantly to ensure nothing overrides it.

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
matthewhannay567: There is a check at the start to see if the robot can start backing up at all, so the values need to be initialed before entering the loop.

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
matthewhannay567: Our robot is differential drive, meaning the only linear velocity is "x", forward or backward

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
matthewhannay567: I don't know dude, it just gets the pose and it works.

igvc_navigation/src/back_up_recovery/back_up_recovery.h
matthewhannay567: The costmap is the map of how much it would cost the robot to go to that area, usually designating areas where the robot would collide with objects.

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
matthewhannay567: This is confusing, but a negative cost means there is a collision, not a positive one.

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
oswinso: We tell it which topic to get the robot pose from, ie. `/odometry/filtered`, and then this is just a method that gets the robot pose from the "base" costmap, which already subscribes to this topic.

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
oswinso: Check [this](http://wiki.ros.org/costmap_2d#Inflation) out for how the costs work

igvc_navigation/src/back_up_recovery/back_up_recovery.h
oswinso: I don't like that `world_model_` and `local_costmap_` are raw pointers and you need to `new` them, since we've got `std::unique_pointer` available. I think you should change these to `std::unique_ptr`

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
oswinso: Instead of repeating this logic twice, I think you should just extract this out into a separate method that you call, maybe something like `bool isNextStepValid()` or something.

oswinso: Check here for why this should be negative:

igvc_navigation/src/back_up_recovery/back_up_recovery.cpp
```diff
@@ -0,0 +1,102 @@
...
+  double current_yaw = tf2::getYaw(global_pose.pose.orientation);
+  geometry_msgs::Point current_position = global_pose.pose.position;
+  geometry_msgs::Point start_position = current_position;
+  bool next_step_valid = nextStepValid(current_position, current_yaw);
+
+  geometry_msgs::Twist vel_msg;
+  vel_msg.linear.x = -velocity_;
+  vel_pub.publish(vel_msg);
+
+  while (n.ok() && igvc::get_distance(current_position, start_position) < threshold_distance_ && next_step_valid)
```
oswinso: `next_step_valid` doesn't get updated anymore while in the loop

Issue #549, opened 2019-10-21
Refactor igvc_perception NN things


oswinso:
[cnn.py](https://github.com/RoboJackets/igvc-software/blob/master/igvc_perception/src/cnn/cnn.py) currently is a bit of a mess and needs to be cleaned up. Also, anything related to the NN in the `igvc_perception` folder is messy.

AC: Refactored cnn.py and `igvc_perception`

suhacker1:
#611 falls under this, right?

oswinso:
Yeah... but I guess like it not conforming to pep8 is only a small part of the "mess". The bigger part of this "mess" is that it's got like multiple responsibilities, and the code itself is structured in a mess and not really the styling.

PR #641, opened 2020-01-26
Created CNN model visualizer node

oswinso:
# Description
Our `cnn.py` node needs a refactor, per #549. Even though that isn't done, (at the cost of code duplication) I think it's important to have something we can use to check how well the model does against bag data.

This PR does the following:
- Creates a `nn_viz` node for visualizing the model's performance on a `/sensor_msgs/CompressedImage` topic for use with visualizing the CNN on bag data
- Updates the `.gitignore` file to reflect the new path to `igvc_sandbox`

# Testing steps (If relevant)
## Testing the visualization
1. Download a CNN model from [cloud.robojackets.org](cloud.robojackets.org)
    a. IGVC -> Software -> model -> IGVCModel_210.pt
2. Put the model in `igvc_sandbox/models`
3. Change the path in `nn_viz.launch` to the correct path
4. Download a competition bag file from [cloud.robojackets.org](cloud.robojackets.org)
5. Run the bagfile with `rosbag play`
6. Run the node using `roslaunch igvc_perception nn_viz.launch`
7. Visualize the results either using `rqt_image_view` or `rosrun rqt_gui rqt_gui`

Expectation: You should see something like this:
Screenshot from 2020-01-26 15-23-30

# Self Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,14 +44,14 @@
 */cmake-build-*/
 
 #Camera Calibration
-/sandbox/camera_pose_estimation/CMakeFiles
-/sandbox/camera_pose_estimation/CMakeCache.txt
-/sandbox/camera_pose_estimation/cmake_install.cmake
-/sandbox/camera_pose_estimation/Makefile
-/sandbox/camera_pose_estimation/solvepnp
+/igvc_sandbox/camera_pose_estimation/CMakeFiles
+/igvc_sandbox/camera_pose_estimation/CMakeCache.txt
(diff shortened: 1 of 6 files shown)


oswinso:
@suhacker1 if you're fine with this, I'm gonna change the line length to 120 because imo 80 is too short

oswinso:
1. Yeah sure. I'll change it to flake8
2. Honestly, I feel like #611 by itself isn't really **that** helpful, as at the end of the day it still needs to be refactored completely as its pretty spaghetti right now. I think it should be fine that we merge this in first.

Issue #629, opened 2020-01-09
Try out traversability analysis for handling ramps


oswinso:
Ramps are a big problem for us right now. We don't have any way of handling it properly. [A long time ago, we used to have an explicit flag for whether we were on the ramp or not, but thats a really dumb way of doing it.](https://github.com/RoboJackets/igvc-software/blob/8eff5894a4338dc1238783135cbba1176b6995cb/2009/candiii/vision/vision.cc#L1480-L1511)

One way that I think would be a good way of handling ramps would be to do traversability analysis. This would probably be done by looking the local gradient of points or something. We probably should do a research and design document for this, as there are many different options and literature surrounding this topic.

Some papers I found:
- Classical approaches (Non nn)
    - [Mobile Robot Traversability Mapping - For Outdoor Navigation](https://pdfs.semanticscholar.org/3b3d/d5aa4cdc8452a6e3730859b0e36a99a23ec3.pdf)
        - Some dude's Master's thesis. Seems quite detailed
    - [Efficient Traversability Analysisfor Mobile Robots using the Kinect Sensor](http://jacoposerafin.com/wp-content/uploads/bogoslavskyi13ecmr.pdf)
        - A more local method, but there's a nice trick for computing gradients that I think we can use
    - [3D traversability awareness for rough
terrain mobile robot](https://www.researchgate.net/publication/263246325_3D_traversability_awareness_for_rough_terrain_mobile_robots)
        - They propose a "descriptor" that can be used to determine traversability. Conveniently includes a ramp example.
    - [Watch Your Step! Terrain Traversability for Robot Control](https://www.intechopen.com/books/robot-control/watch-your-step-terrain-traversability-for-robot-control)
        - An overview paper. Includes the approach above.
    - [`elevation_mapping`](https://github.com/ANYbotics/elevation_mapping)
        - More of a tool than how to do traversability analysis, but I think we should look into using this as it does some nice Bayesian inference for calculating the height of each cell
- NN approaches (I'm less confident about as I think data collection will be time consuming and I don't know how robust this is)
    - [Traversability Analysis for Mobile Robots in Outdoor Environments:A Semi-Supervised Learning Approach Based on 3D-Lidar Data](http://ais.informatik.uni-freiburg.de/publications/papers/suger15icra.pdf)
[...]

PR #642, opened 2020-01-30
Added design document for slope detection.

matthewhannay567:
# Description
Adds design document for slope detection.

Related to #629


oswinso:
@matthewhannay567 
I think you should assign another person to review this.

**Before** you do that though, I think since this is like your 2nd / 3rd PR or something and this is like an actual hard area that you should switch away from using the "new member project design document template" and do something like [FirmwareRefactor.md](https://github.com/RoboJackets/igvc-software/blob/master/documents/design/FirmwareRefactor.md) or [mapping_refactor.md](https://github.com/RoboJackets/igvc-software/blob/master/documents/design/mapping_refactor.md).

Also, given that this isn't a new member project, I think you should do some research and testing first before writing up the design document, ie.
- What are existing approaches to doing traversability analysis?
- How do you do height mapping?
- How does the math behind ANYBOTICS `elevation_mapping` work?
- How well does the package work?
- How do you calculate the slope? (Maybe computationally efficient approaches or something)
- Should we use ANYBOTICS `elevation_mapping` or write our own solution?

I'll make an issue for it, but basically make a research document first, like [barrel_detection](https://github.com/RoboJackets/igvc-software/blob/master/documents/research/barrel_detection.md) or something. After you have answered most of the questions above, **then** write a design document outlining the solution you decide to go with. I think given how this area is a bit more theory-heavy it would be nice to do a small presentation on it after you're done with the design document.

documents/design/slope_detection.md
```diff
@@ -0,0 +1,35 @@
...
+2. Add the points to an elevation map.
+3. Find the gradient at every point in the elevation map.
+4. Generate a costmap based on that gradient.
+
+## Design
+
+`pointcloud_filter` will filter out any points that are far away or behind the robot.
+
+With the `elevation_mapping` library, we can make a grid map with elevation in each cell. A node is created that
+subscribes to the filtered pointcloud and publishes an elevation map.
```
oswinso: In addition to talking about what the `elevation_mapping` does, I think it would be nice to
1. link to the related papers
2. add a link to the library on github
3. talk about how it is doing elevation mapping (ie. give a brief overview of the paper)
4. talk about why we chose to use this, and maybe some alternatives that were also considered.

PR #652, opened 2020-02-20
add design doc

### LINKED ISSUE #651: Actual Barrel Detection With Neural Network
A more robust solution compared to the Clustering and RANSAC solution. Basically treat LIDAR data like images and feed it to a neural network. A related paper could be found here: https://arxiv.org/abs/1612.00593

FZ2000:
# Description

Add design doc for a Neural Network based barrel segmentation algorithm. 

This PR does the following:
- Add design doc

Fixes #651

### DIFF
--- a/documents/design/actual_barrel_segmentation-cnn.md
+++ b/documents/design/actual_barrel_segmentation-cnn.md
@@ -1,0 +1,18 @@
+# Actual Barrel Detection (Neural Network) Design Document
+
+**Author: Zhiyan Zhou (Frank)**
+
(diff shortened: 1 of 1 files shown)


matthewhannay567:
Some questions:

- Where are you going to get the training data for the CNN?  How are you going to train this CNN?
- What exactly is the CNN going to output?
- How will the CNN handle multiple barrels in the same image? A barrel that is slightly obfuscated?

matthewhannay567:
Can we get an update?

PR #672, opened 2020-03-01
Traversability Costmap Implementation

### LINKED ISSUE #629: Try out traversability analysis for handling ramps

### LINKED ISSUE #664: traversability_layer for integrating the traversability analysis into the navigation stack
While #629 performs the traversability analysis, we need a way to incorporate that into the current mapping framework by creating a `Costmap2DLayer`. By doing this, we will deprecate `lidar_layer`.

AC: `traversability_layer` that results in the surface of the ramp not showing as an obstacle, but the sides of the ramp are an obstacle in simulation.

matthewhannay567:
# Description

This PR adds a traversability_layer costmap plugin that uses elevation mapping and slope calculation to make a costmap based on the slope of the terrain.

This requires downloading the `kindr`, `kindr_ros` and `elevation_mapping` libraries.

This PR does the following:
- adds elevation mapping
- adds traversability costmap layer
- makes move_base_flex use traversability_layer and not lidar_layer

Fixes #629, Fixes #664 
# Testing steps (If relevant)
## Test Case 1
1. Go into a simulation.
2. Put the robot in front of a ramp.

Expectation: The robot should not see the ramp as an obstacle, but the edges of the ramp should be detected as an obstacle. Barrels should also be seen as an obstacle.

# Self Checklist


oswinso:
Make sure to merge in the changes from #676 to fix the build on circle ci

PR #678, opened 2020-03-05
IMU Improvements and Research Document

### LINKED ISSUE #633: Characterize magnetometer behavior in current IMU
We still don't know too much about what's happening with the magnetometer in our IMU.

After calibration, it seems like the it takes time for the heading to "converge" to a value, but that value seems(?) to be different or something.

We need to actually figure out our magnetometer and understand exactly what is happening & why.

One thing I think would be very helpful is to plot of the distribution over time of the magnetometer heading from start up to convergence, ie. do multiple runs of recording the heading that the IMU gives us, waiting for like 2 - 3 minutes, then unplugging the IMU and repeating.

Also, doing the same plot with the raw heading (which we don't use), and comparing it with the filtered heading (the one we're currently using) to get more information about which one we should use.

AC: A research document on the current magnetometer, with full descriptions of the types of behaviour for the magnetometer that we are seeing right now.

tangemicioglu:
# Description

The IMU has been having problems right after sensors are launched, suspicion was placed on the magnetometer but the cause appears to be the IMU's filtering. This PR gives more tools to control the values and observe them, and adds a research document explaining results.

This PR does the following:
- Requests magnetometer data and publishes it in addition to IMU message.
- Adds gyro calibration to launchfiles, only calibrates when explicitly requested.
- Adds research document detailing information about IMU problems.

Fixes #633 

# Testing steps (If relevant)
## Test Case 1
1. Connect the top IMU to the robot and launch `imu_top.launch`. Do not move the robot during the next steps.
2. Launch `rqt_plot` and add `/magnetometer/mag` to visualization. It should be noisy but stable.
3. Add `/magnetometer/orientation` to visualization. Observe for at least one minute. It should gradually converge to a value or be stable depending on when it's launched.
4. Relaunch `imu_top.launch`, repeat 3 multiple times. They should converge to same value.

Expectation: The IMU publishes magnetic field in addition to the IMU data, gyro doesn't recalibrate unless requested and convergence is consistent.

# Self Checklist


igvc_platform/src/imu/YostLabDriver.cpp
```diff
@@ -1,37 +1,80 @@
...
-  // use identity matrix as default orientation correction
+  this->magnet_pub_ = this->yostlab_nh_.advertise<sensor_msgs::MagneticField>("/imu_mag", 10);
+  this->updater.setHardwareIDf("IMU: %s", this->getSerialPort().c_str());
+  this->updater.add("IMU Diagnostic", this, &YostLabDriver::imu_diagnostic);
 
+  // use identity matrix as default orientation correction
   assertions::param(this->yostlab_priv_nh_, "imu_orientation_correction", this->imu_orientation_correction_,
                     std::vector<double>{ 1, 0, 0, 0, 1, 0, 0, 0, 1 });
   assertions::param(yostlab_priv_nh_, "orientation_rotation", orientation_rotation_, 0.0);
+  assertions::param(yostlab_priv_nh_, "calibrate_gyro", calibrate_gyro_, false);
```
andrewyarovoi: needs to be consistent with imu_top.launch parameter name

igvc_platform/launch/sensors.launch
```diff
@@ -22,10 +22,14 @@
     <include file="$(find igvc_platform)/launch/velodyne_lidar.launch" />
 
     <!-- Bottom IMU -->
-    <include file="$(find igvc_platform)/launch/imu_bottom.launch" />
+    <include file="$(find igvc_platform)/launch/imu_bottom.launch" >
+        <arg name="calibrate_imu" value="false"/>
```
andrewyarovoi: The argument "calibrate_imu" is already set to false in the imu_bottom.launch. Adding it again is kind of unnessary repetition. I think it would be better to have it only in one spot, so if it is changed, the values actually take effect.

igvc_platform/src/imu/YostLabDriver.cpp
```diff
@@ -168,16 +170,20 @@ std::string YostLabDriver::getMIMode()
     else
       return "Unknown";
   }();
-  ROS_INFO_STREAM(logger << "MI Mode: " << ret_buf << ", buf is: " << buf);
+  ROS_INFO_STREAM(this->logger << "MI Mode: " << ret_buf << ", buf is: " << buf);
```
andrewyarovoi: I'm only saying it because I'm scared of what Oswin would say if he saw `this` again, but I don't think it's necessary.

igvc_platform/launch/sensors.launch
```diff
@@ -22,10 +22,14 @@
...
+    </include>
 
     <!-- Top IMU -->
-    <include file="$(find igvc_platform)/launch/imu_top.launch" />
+    <include file="$(find igvc_platform)/launch/imu_top.launch" >
+        <arg name="calibrate_imu" value="false"/>
```
andrewyarovoi: Same issue as above. It's the variable is already in the launch file.

igvc_platform/launch/sensors.launch
tangemicioglu: This is intentional, the default value in `imu_bottom` and `imu_top` shouldn't be changed, but we may have to set the value to true in `sensors.launch` if we decide to recalibrate every time. This allows quick changes and maintains a more object-like structure. `simulation.launch` follows a similar structure.

igvc_platform/src/imu/YostLabDriver.h
```diff
@@ -138,7 +145,7 @@ class YostLabDriver : SerialInterface
   Slot #3: corrected acceleration vector [3x float]
   Slot #[4-8]: No Command
```
oswinso: Does this need to be updated?

igvc_platform/src/imu/YostLabDriver.cpp
```diff
@@ -200,7 +206,7 @@ void YostLabDriver::run()
       if (terms_count == 1)
       {
         // Verify that it is a complete message of 4 lines
-        if (line_num_ == 4)
+        if (line_num_ == 5)
```
oswinso: Number doesn't match comment - update the comment

PR #684, opened 2020-03-06
changed how global path is replanned

matthewhannay567:
# Description

Fixes an issue where `navigation_server` would sometimes keep path planning even when the goal was reached.

This PR does the following:
- Changes the function call for runGetPath() to be at the end of actionGetPathDone() with a timer and mutex lock. This is more how like move_base_flex did it.

# Testing steps (If relevant)
## Test Case 1
1. open a simulation
2. run `navigation_simulation.launch` with `read_from_file:=false`
3. open rviz
4. set a goal close to the robot

Expectation: When the robot reaches the goal, it will not attempt to plan again. This was a rare occurrence under normal circumstances, so multiple goals should be sent.

# Self Checklist

### DIFF
--- a/igvc_navigation/src/navigation_server/navigation_server.cpp
+++ b/igvc_navigation/src/navigation_server/navigation_server.cpp
@@ -7,15 +7,16 @@ NavigationServer::NavigationServer()
   , recovery_trigger_(NONE)
   , nh_()
   , private_nh_("~")
-  , action_client_get_path_(private_nh_, "/move_base_flex/get_path")
-  , action_client_exe_path_(private_nh_, "/move_base_flex/exe_path")
-  , action_client_recovery_(private_nh_, "/move_base_flex/recovery")
+  , action_client_get_path_(private_nh_, "/move_base_flex/get_path", true)
+  , action_client_exe_path_(private_nh_, "/move_base_flex/exe_path", true)
+  , action_client_recovery_(private_nh_, "/move_base_flex/recovery", true)
   , action_server_(private_nh_, "/move_to_waypoint", boost::bind(&NavigationServer::start, this, _1),
                    boost::bind(&NavigationServer::cancel, this), false)
 {
   ROS_INFO_STREAM_NAMED("nav_server", "Navigation server created.");
   assertions::getParam(private_nh_, "recovery_enabled", recovery_enabled_);
   assertions::getParam(private_nh_, "oscillation_distance", oscillation_distance_);
+  assertions::getParam(private_nh_, "max_replanning_tries", max_replanning_tries_);
 
   double timeout, wait_time, rate;
   assertions::getParam(private_nh_, "oscillation_timeout", timeout);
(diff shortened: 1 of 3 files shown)


igvc_navigation/src/navigation_server/navigation_server.cpp
```diff
@@ -147,35 +150,82 @@ void NavigationServer::actionGetPathDone(const actionlib::SimpleClientGoalState
...
+      size_t size = path.poses.size();
+      // change the orientation of the last pose to that of the pose before it
+      path.poses[size - 1].pose.orientation = path.poses[size - 2].pose.orientation;
+    }
+    runExePath(path);
+  }
+
+  replanning_mtx_.lock();
+  replanning_rate_.sleep();
+  replanning_mtx_.unlock();
```
oswinso: Rather than do this I think you should create a new scope with a guard in it, ie.
```C++
{
  std::lock_guard<std::mutex> guard{replanning_mtx_};
  replanning_rate_.sleep();
}
```

igvc_navigation/src/navigation_server/navigation_server.h
```diff
@@ -64,8 +64,8 @@ class NavigationServer
 
   ros::Time start_time_;
 
-  ros::Time time_of_last_get_path_;
-  ros::Duration time_between_get_path_ = ros::Duration(0.5);
+  ros::Rate replanning_rate_ = ros::Rate(1.0);
+  boost::mutex replanning_mtx_;
```
oswinso: The standard library already has a `std::mutex`, theres no need to use `boost::mutex`

oswinso:
I think you should explain a bit more (so its somewhere on github) why adding this mutex fixes the issues you were describing.

It isn't immediately obvious why the problem was happening, and how adding the mutex fixes the issue.

igvc_navigation/src/navigation_server/navigation_server.cpp
```diff
@@ -173,52 +174,82 @@ void NavigationServer::actionGetPathDone(const actionlib::SimpleClientGoalState
       break;
   }
 
-  if (navigation_state_ != EXE_PATH)
+  // only start replanning if we are executing path
+  if (navigation_state_ == EXE_PATH)
   {
-    return;
+    current_replanning_tries_ = 0;
+    // prevent other threads from replanning while thread sleeps
```
oswinso: Why are there multiple threads trying to replan?

matthewhannay567: I think the reason they have the mutex is because there two points where replanning is done: are the end of actionGetPathDone and actionGetPathReplanningDone. The mutex prevents both from executing at the same time, which could happen if the robot finishing replanning just as it gets a new navigation goal.

igvc_navigation/src/navigation_server/navigation_server.cpp
```diff
@@ -147,35 +151,115 @@ void NavigationServer::actionGetPathDone(const actionlib::SimpleClientGoalState
...
+  }
+}
+
+void NavigationServer::actionGetPathReplanningDone(const actionlib::SimpleClientGoalState &state,
+                                                   const mbf_msgs::GetPathResultConstPtr &result_ptr)
+{
+  // check if we are executing path
+  if (navigation_state_ == EXE_PATH)
+  {
+    const mbf_msgs::GetPathResult &result = *(result_ptr.get());
```
oswinso: `GetPathResultConstPtr` is a `boost::shared_ptr`, so you should be able to just dereference it without doing `*(result_ptr.get())`

PR #688, opened 2020-03-13
Costmap Layers Refactor

### LINKED ISSUE #655: Messy implementation of local / global costmaps
Currently, our costmap layers looks like this:
**Global**:
- `unrolling_layer`

**Local**
- `lidar_layer`
- `line_layer`

Originally, it was supposed to be
**Global**:
- `lidar_layer`
- `line_layer`

**Local**
- `static_layer`

However, this didn't work, since [`static_layer` blocks in the constructor until it receives the map at least once](https://github.com/ros-planning/navigation/blob/99ac7744e9ddb98048fafcbf65e8972153936202/costmap_2d/plugins/static_layer.cpp#L90-L99), but `move_base_flex` initializes the local costmap first and then the global costmap first, meaning that the `static_layer` is just stuck spinning.

In order to "fix" this, I tried moving the `lidar_layer` and `line_layer` to the local costmap while having a `unrolling_layer` for the global costmap, which basically "stitches together" the local costmap (which is rolling in order to reduce the size so that TEB plans faster).

This is a bad approach though and results in a variety of issues (ie. "forgetting" when you reenter an area that you previously left). The better way would be to have our own version of `static_layer` that doesn't block in the constructor, so that the global planner can get initialized and start publishing maps.

AC: `line_layer` and `lidar_layer` on the global costmap (with `rolling_costmap` False since its now on the global costmap), and then having the local costmap read small subsections from that global costmap (with `rolling_costmap` True) by fixing the issue described earlier, ie. creating our own version of `static_layer` that doesn't block in the constructor.

matthewhannay567:
# Description
This PR rearranges the costmap layers so that the sensor data is applied directly to the global costmap and the local costmap is a zoom in on the global costmap, instead of the local costmap "constructing" the global costmap

This PR does the following:
- changes the global costmap plugins to `traversability_layer`, `line_layer`, and `inflation_layer`
- changes the local costmap plugins to `rolling_layer`
- creates `rolling_layer`, which copies data from the global costmap that surrounds the robot

Fixes #655 

# Testing steps (If relevant)
## Test Case 1
1. Run a simulation
2. Open rviz
3. Set a goal

Expectation: The global costmap should update according to `traversability_layer` and `line_layer`. The local costmap should copy over the LETHAL cost cells.

# Self Checklist


oswinso:
Not sure if introduced by this PR, but something seems to be dying:
```
[ERROR] [1585168957.160760913, 1739.211000000]: Could not get pose information from robot for time 1739.133500. Buffer empty?
[ERROR] [1585168957.160779640, 1739.211000000]: Updating process noise failed.
[move_base_flex-5] process has died [pid 27780, exit code -11, cmd /home/oswinso/catkin_ws/devel/lib/mbf_costmap_nav/mbf_costmap_nav __name:=move_base_flex __log:=/home/oswinso/.ros/log/2481a9cc-6ed4-11ea-80a7-7085c28c1dab/move_base_flex-5.log].
log file: /home/oswinso/.ros/log/2481a9cc-6ed4-11ea-80a7-7085c28c1dab/move_base_flex-5*.log
```

tangemicioglu:
I fixed my dependency issues and I only get the error Oswin mentioned when I restart navigation while the robot is moving, so I don't think it's anything to worry about.

PR #693, opened 2020-03-16
Small gazebo world to test ramp

oswinso:
# Description
Our gazebo worlds are all currently very large, and the ramp is always very far away.

This PR does the following:
- Creates a gazebo world where the ramp is literally right in front of the robot

# Testing steps (If relevant)
## Run the gazebo world
1. `roslaunch igvc_gazebo ramp_lane.launch`
2. `roslaunch igvc_navigation navigation_simulation.launch`

Expectation: There's a ramp right in front and the stack doesn't break.

# Self Checklist


oswinso:
Yeah. Keep in mind this is was still with @jviszlai's ground plane filtering and not @matthewhannay567's traversability though

PR #706, opened 2020-03-29
Feat/elevation map height node

matthewhannay567:
# Description

This PR does the following:
- Adds a node which takes in `odometry/filtered` and changes the z value and z variance to data gained from elevation_mapping.
- Adds `elevation_map_height.launch`, which launches the node and has a parameter for the output topic.

# Testing steps
1. Launch qualification.launch
2. Launch navigation_simulation.launch read_from_file:=false
3. Launch elevation_map_height.launch
4. rostopic echo `/odometry/elevation_map_estimation`

Expectation: The pose's z value will be nan, because there is no elevation map data below the robot.

5. Launch rviz
6. Send a goal so the robot ends up above elevation map data (going forward quite a bit will work)

Expectation: The pose's z value will be an actual number, equal to the elevation map height at that point.

# Self Checklist


igvc_perception/src/elevation_map_height_node/elevation_map_height_node.h
```diff
@@ -0,0 +1,26 @@
+#ifndef SRC_ELEVATIONMAPHEIGHTNODE_H
+#define SRC_ELEVATIONMAPHEIGHTNODE_H
```
VAM7686, line 2: do

VAM7686, line 2: Do you need to have this import when you already have the line above?

matthewhannay567: These lines are there to ensure that `elevation_map_height_node.h` is only initialized once

matthewhannay567:
> 
> 
> Everything works! One question about getting the elevation: why do you use elevation_smooth layer instead of elevation layer? Does elevation smooth just smooth out the grid map elevations?

Yes, `elevation_smooth` smooths and fills gaps in the height map.

PR #787, opened 2020-10-31
Feat/add required to differential drive

averylg:
# Description

{{ Setting the nodes in differential_drive.launch to be required }}

This PR does the following:
- Adds the attribute `required="true"` to the node tag in differential_drive.launch

Fixes #{{ Adds `required="true"` attribute in the launch file }}

Expectation: Sets nodes in differential_drive.launch to be required


matthewhannay567 (changes requested):
Change the PR base from `master` to `noetic-migration`

PR #791, opened 2020-10-31
Feat/add required to pointcloud filter

averylg:
# Description

{{ First issues: pointcloud filter }}

This PR does the following:
- Adds the field `required="true"` to the `pointcloud_filter` node tag in the pointcloud_filter.launch file

Fixes #{{ Node previously not set to required in pointcloud_filter.launch }}

Expectation: `pointcloud_filter` node set to required


matthewhannay567 (approved):
Looks good!

PR #801, opened 2020-11-01
Feat/add required to elevation map visualization.launch

prsingareddy:
# Description

Added attribute required="true" to the following nodes elevation_map_fused_visualization and elevation_map_raw_visualization
in elevation_map_visualization.launch.


matthewhannay567 (approved):
Looks good!

PR #802, opened 2020-11-01
Feat/add required to elevation map height.launch

prsingareddy:
# Description

Added attribute required="true" to node elevation_map_height_node in elevation_map_height.launch.


matthewhannay567 (changes requested):
You should make sure that your branches are up to date with `noetic-migration`, so you should merge the latest `noetic-migration` branch into your feature branch. Because right now there are some files being changed that shouldn't be.

matthewhannay567 (approved):
Looks good to me!

PR #808, opened 2020-11-05
Feat/add required to igvc control

### LINKED ISSUE #756: Set nodes in igvc_control.launch to be required
For the following node tags in `igvc_control.launch`, found in `igvc_gazebo`, add the attribute `required="true"`: 
* igvc_control

Once you make the PR, assign a fellow RoboNav member to review it, be it an old or new member.

NicholasLeone:
This PR sets the igvc_control node to have required "true"

Fixes #756 

# Testing steps (If relevant)
1. Run igvc_control.launch
2. Run rosnode kill on igvc_control

Expectation: The entire stack dies when the igvc_control node dies.
# Self Checklist


matthewhannay567 (changes requested):
This is merging to `master` instead of `noetic-migration`, and when I try to change the base it says there is not difference between `noetic-migration` and `feat/add_required_to_igvc_control`, so I think you forgot to add the `required` attribute to the node.

PR #811, opened 2020-11-15
Feat/scan to pointcloud node cpp

averylg:
# Description

Separates the scan_to_pointcloud node into a header file and a C++ class.

# Testing steps (If relevant)
1. Run simulation on noetic-migration branch
2. Run simulation on feat/scan_to_pointcloud_node_cpp branch
3. Check that the results for the scan_to_pointcloud node are the same for both branches.

Expectation: The scan_to_pointcloud node should maintain functionality

# Self Checklist
- [ ] I have tested that the new behavior works


matthewhannay567 (changes requested):
There should be a separate .h and .cpp file, where the header defines the member variables and callback functions, and the cpp file defines the callback functions. The `main` function should be just ros::init, the class constructor, then ros::spin. Everything else gets moved to the constructor.

matthewhannay567:
Also, make sure to run `make format` on the code before pushing, to ensure that the code is formatted correctly.

PR #822, opened 2021-01-12
Remove respawn=true so that simulations run

### LINKED ISSUE #821: Simulations error on startup
Several of the launch files involved in launching jessi simulations have both a respawn=true and a required=true tag, and thus error on launch. Remove the respawn=true tag so the simulations run.

andrewyarovoi:
# Description
Fixes a mistake that was previously merged in. This PR removes the respawn=true tag from several nodes so that both simulation.launch and simulation_low.launch actually work. 

This PR does the following:
- Fixes the simulations so they launch.

Fixes #821 

# Testing steps (If relevant)
## Test Case 1
1. run `catkin_make`
2. run `roslaunch igvc_gazebo autonav_low.launch`
3. check that gazebo opens
4. close gazebo and run `roslaunch igvc_gazebo qualification.launch`
5. check that gazebo opens

Expectation: Both the regular and the low simulations open and run normally.

# Self Checklist

### DIFF
--- a/igvc_gazebo/launch/simulation.launch
+++ b/igvc_gazebo/launch/simulation.launch
@@ -59,7 +59,7 @@
         <param name="mag_field_variance" value = "0.000001"/>
     </node>
 
-    <node name="ground_truth_to_rpy" pkg="igvc_utils" type="quaternion_to_rpy" respawn="true" output="screen" required="true">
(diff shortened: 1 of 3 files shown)


matthewhannay567 (approved):
Looks good

PR #827, opened 2021-02-21
Splits control node into a c++ class and header

### LINKED ISSUE #738: Rewrite the igvc_gazebo/control node to be a C++ class.
AC: The igvc_gazebo/control node should be split into a header and cpp file, with the cpp file containing member function implementations as well as the main function, and retain functionality of the node.

navaneet614:
Makes the control node a c++ class with header

Fixes #738 

# Testing steps (If relevant)
## Test Case 1
1. Test original node
2. Test new node on this branch

Expectation: Everything still works fine

# Self Checklist

### DIFF
--- a/igvc_gazebo/nodes/control/control.cpp
+++ b/igvc_gazebo/nodes/control/control.cpp
@@ -1,3 +1,5 @@
+#include "control.h"
+
 #include <igvc_msgs/velocity_pair.h>
 #include <parameter_assertions/assertions.h>
 #include <ros/ros.h>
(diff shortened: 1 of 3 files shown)


igvc_gazebo/nodes/control/control.cpp
```diff
@@ -171,6 +136,48 @@ int main(int argc, char **argv)
...
+  }
+
+  // ROS_INFO_STREAM("right speed: " << speed_measured_right << " left speed: " << speed_measured_left);
+}
+
+int main(int argc, char **argv)
+{
+  ros::init(argc, argv, "jessii_controller");
+
+  Control c;
```
VAM7686, line 180: Currently `c` is uninitialized. Make sure to add a constructor otherwise nothing will happen.

VAM7686, line 180: Currently `c` is uninitialized. Make sure to add a constructor or nothing will happen.

igvc_gazebo/nodes/control/control.cpp
```diff
@@ -1,3 +1,5 @@
+#include "control.h"
+
 #include <igvc_msgs/velocity_pair.h>
 #include <parameter_assertions/assertions.h>
 #include <ros/ros.h>
```
VAM7686, line 5: Make sure to put your includes in the header file. You define `<igvc_msgs/velocity_pair.h>` and `<sensor_msgs/JointState.h>` twice.

VAM7686 (changes requested):
Make sure to merge master into this branch to fix the CI build.

Also, add some more descriptive testing steps. Testing steps should be reproducible and expectation of the test case should be the expected result of the testing steps.

PR #830, opened 2021-06-05
Back Circle Service

### LINKED ISSUE #749: Convert back_circle into a ROS service
Currently at the start of the autonav course, MoveBaseFlex gives us this path: 
image

This is a problem as it takes a while for the robot to start going pathing the correct way, and the robot sometimes gets stuck while turning around. The solution to this is to add a layer to the costmap which consists of a half-circle blocking the backwards path.

We currently have [this code](https://github.com/RoboJackets/igvc-software/blob/master/igvc_navigation/src/back_circle/main.cpp) which generates a pointcloud message representing the back circle. However right now it is doing this constantly.

We want to to be a ROS service, so it can called only when needed: when navigation_client sends a waypoint goal.

AC:
- back_circle converted to a ROS service
- back_circle service called every time navigation_client sends waypoint to navigation_server

matthewhannay567:
# Description
Adds a layer to the costmap that prevents the robot from pathing backwards.

Fixes #749 

## Test Case 1
1. Start up gazebo simulation
2. Launch navigation with read_from_file:=false
3. Launch rviz and set a navigation waypoint
4. Set another one later

Expectation: The back circle should move with the robot and always generate behind it when setting waypoints.

## Test Case 2
1. Start up gazebo simulation
2. Launch navigation with read_form_file:=true

Expectation: The back circle will be generated behind the robot once it reads from the waypoint file.

# Self Checklist


igvc_navigation/src/back_circle/back_circle_server.cpp
```diff
@@ -0,0 +1,84 @@
...
+  std::vector<double> xVec, yVec, zVec;
+
+  assertions::getParam(pNh, std::string("width"), width);
+  assertions::getParam(pNh, std::string("length"), length);
+  assertions::getParam(pNh, std::string("grid_size"), grid_size);
+  assertions::getParam(pNh, std::string("thickness"), thickness);
+  assertions::getParam(pNh, std::string("offset"), offset);
+  double meters = (length + width) * 2;
+
+  using namespace cv;
```
VAM7686, line 34: Try to avoid this

VAM7686, line 34: Try to avoid doing this.

VAM7686 (changes requested):

igvc_navigation/src/back_circle/back_circle_server.cpp
VAM7686: Fixed.

VAM7686 (approved):
lgtm

Issue #750, opened 2020-09-30
Create costmap layer that uses back_circle pointcloud


matthewhannay567:
Continuing from #749 

Now that we have the pointcloud data from back_circle, we need a costmap layer plugin that uses it. Then that costmap needs to be added to the global_costmap_params file. 

AC:
- pointcloud data is converted to costmap
- back circle costmap layer is added to global costmap.

PR #835, opened 2021-07-29
Back Circle

### LINKED ISSUE #749: Convert back_circle into a ROS service

VAM7686:
# Description
Adds a layer to the costmap that prevents the robot from pathing backwards.

Fixes #749, #750 

## Test Case 1
1. Start up gazebo simulation
2. Launch navigation with read_from_file:=false
3. Launch rviz and set a navigation waypoint
4. Set another one later

Expectation: The back circle should move with the robot and always generate behind it when setting waypoints.

## Test Case 2
1. Start up gazebo simulation
2. Launch navigation with read_form_file:=true

Expectation: The back circle will be generated behind the robot once it reads from the waypoint file.

## Back Circle in autonav.launch:
back circle in rviz

# Self Checklist


igvc_navigation/src/back_circle/back_circle_server.cpp
```diff
@@ -0,0 +1,82 @@
...
+  double meters = (length + width) * 2;
+
+  int img_size = static_cast<int>(std::round(meters / grid_size));
+  cv::Mat img(img_size, img_size, CV_8U, cv::Scalar::all(0));
+  cv::ellipse(img, cv::Point(img_size / 2, img_size / 2), cv::Size(width / grid_size, length / grid_size), 0, 90, 270,
+              cv::Scalar(255), static_cast<int>(std::round(thickness / grid_size)));
+
+  pcl::PointCloud<pcl::PointXYZ> back_circle_pointcloud_base_footprint;
+  back_circle_pointcloud_base_footprint.header.stamp = ros::Time::now().toNSec() / 1000;
+  back_circle_pointcloud_base_footprint.header.frame_id = "/base_footprint";
```
VAM7686, line 41: Why is does a slash prefix `base_footprint` here but elsewhere there no slashes?

igvc_navigation/src/back_circle/back_circle_server.h
```diff
@@ -0,0 +1,28 @@
...
+{
+public:
+  BackCircleServer();
+  ros::NodeHandle nh_;
+
+private:
+  void waitForTransform();
+  bool backCircleCallback(igvc_msgs::BackCircle::Request &req, igvc_msgs::BackCircle::Response &res);
+  ros::ServiceServer back_circle_service_server_;
+  tf::TransformListener tf_listener_;
```
VAM7686, line 25: Should we be using tf2 here?

igvc_navigation/launch/back_circle_server.launch
```diff
@@ -0,0 +1,9 @@
+<launch>
+    <node name="back_circle" pkg="igvc_navigation" type="back_circle" output="screen" required="true">
+        <param name="width" value="1"/>
+        <param name="length" value="1.4"/>
+        <param name="grid_size" value="0.1"/>
+        <param name="thickness" value="0.2"/>
+        <param name="offset" value="-0.5"/>
```
VAM7686, line 7: Could we set these parameters in a yaml file like `global_costmap_params.yaml` instead of here?

VAM7686 (commented):
Looks good! Just have a few questions.

igvc_msgs/srv/BackCircle.srv
```diff
@@ -0,0 +1,5 @@
+geometry_msgs/PoseStamped pose # Orientation and position of the robot
```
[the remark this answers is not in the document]
matthewhannay567: I'll remove that.

igvc_navigation/src/navigation_client/navigation_client.cpp
```diff
@@ -50,6 +54,11 @@ void NavigationClient::waitForServer()
     ROS_INFO_STREAM("Waiting for the navigation server to come up");
   }
   ROS_INFO_STREAM("Connected to navigation server!");
+  while (!back_circle_client.waitForExistence(ros::Duration(waiting_time)))
+  {
+    ROS_INFO_STREAM("Waiting for the back circle service server to come up");
```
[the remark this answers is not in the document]
matthewhannay567: Fixed.

igvc_navigation/src/navigation_client/navigation_client.cpp
```diff
@@ -24,6 +27,7 @@ NavigationClient::NavigationClient()
   {
     std::vector<geometry_msgs::PointStamped> waypoints = loadWaypointsFromFile();
     ROS_INFO_STREAM("Reading from file.");
+    ros::Duration(2.0).sleep();
```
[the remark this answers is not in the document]
matthewhannay567: It gives it extra time to read the waypoints.

igvc_navigation/src/back_circle/back_circle_server.cpp
matthewhannay567: It doesn't really matter, and looking at the file it's in it also uses the slash prefix.

igvc_navigation/src/back_circle/back_circle_server.h
matthewhannay567: Ripping out tf stuff to replace it with tf2 is something that I don't think is necessary right here.

igvc_navigation/launch/back_circle_server.launch
matthewhannay567: I'll make it so the config class handles this.

matthewhannay567: Actually because the global_costmap_params.yaml file controls the layers and not the BackCircleServer, it can't be done.

NicholasLeone (approved):
In both test cases, a back circle is successfully generated whenever a new waypoint is placed. 

However, I noticed that when placing a waypoint behind the robot, the back circle forces the robot to find a path around the back circle, rather than turning around and move towards the waypoint.




https://www.dropbox.com/s/ojuhxfdf9r7144i/kazam_28ay5oe6.movie.mp4?dl=0

igvc_navigation/src/mapper/back_circle_layer.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+PLUGINLIB_EXPORT_CLASS(back_circle_layer::BackCircleLayer, costmap_2d::Layer)
+
+namespace back_circle_layer
+{
+BackCircleLayer::BackCircleLayer() : private_nh_("~")
+{
+}
+
+// Get a subscriber in here that subscribes to a topic published to in the main.cpp
```
NicholasLeone, line 12:
```suggestion
// Get a subscriber in here that subscribes to a topic published to in the back_circle_server.cpp
```

igvc_navigation/src/mapper/back_circle_layer.cpp
```diff
@@ -0,0 +1,63 @@
...
+  *max_x = getOriginX() + getSizeInMetersX();
+  *min_y = getOriginY();
+  *max_y = getOriginY() + getSizeInMetersY();
+}
+
+void BackCircleLayer::updateCosts(costmap_2d::Costmap2D &master_grid, int min_i, int min_j, int max_i, int max_j)
+{
+  uint8_t *master_array = master_grid.getCharMap();
+  uint8_t *line_array = getCharMap();
+  unsigned int span = master_grid.getSizeInCellsX();
```
NicholasLeone, line 36: What are these variables used for?

igvc_msgs/srv/BackCircle.srv
```diff
@@ -0,0 +1,4 @@
+---
+float64[] x
+float64[] y
+float64[] z
```
NicholasLeone, line 4: Do we actually use the z vector when inserting values into the costmap?

igvc_navigation/src/mapper/back_circle_layer.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+  for (size_t i = 0; i < xVals.size(); ++i)
+  {
+    // If back_circle_pointcloud contains the point, add it to the costmap
+    // might have to do tf transforms
```
NicholasLeone, line 41: We do not use tf transform here. 
```suggestion
    // If costmap contains the back circle point, add it to the costmap
```

NicholasLeone (changes requested):
Looks good to me. Just a few changes.

matthewhannay567 (approved):
works

PR #861, opened 2021-12-07
Ground truth

### LINKED ISSUE #739: Rewrite the igvc_gazebo/ground_truth node to be a C++ class
AC: The igvc_gazebo/ground_truth node should be split into a header and cpp file, with the cpp file containing member function implementations as well as the main function, and retain functionality of the node.

nzhongathan:
# Description

Takes the main.cpp file for the ground_truth node and splits it into a .cpp and header file for a ground truth republished node.

This PR does the following:
- Creates a header file for the ground truth node
- Creates the ground truth node

Fixes #739 

# Testing steps (If relevant)
1. Ground truth republisher still works properly

# Self Checklist
- [x] I have tested that the new behavior works

### DIFF
--- a/igvc_gazebo/nodes/ground_truth/ground_truth_republisher.cpp
+++ b/igvc_gazebo/nodes/ground_truth/ground_truth_republisher.cpp
@@ -17,9 +17,9 @@ GroundTruthRepublisher::GroundTruthRepublisher()
 	assertions::param(pNh, "longitude", longitude, -84.405001);
 	assertions::param(pNh, "latitude", latitude, 33.774497);
 
-	ros::Subscriber ground_truth = nh.subscribe<nav_msgs::Odometry>(ground_truth_topic, 10, groundTruthCallback);
+	ros::Subscriber ground_truth = nh.subscribe<nav_msgs::Odometry>(ground_truth_topic, 10, &GroundTruthRepublisher::groundTruthCallback, this);
 
-	ros::Subscriber estimate_sub = nh.subscribe<nav_msgs::Odometry>("/odometry/filtered", 1, odomCallback);
+	ros::Subscriber estimate_sub = nh.subscribe<nav_msgs::Odometry>("/odometry/filtered", 1, &GroundTruthRepubisher::odomCallback, this);
 	g_ground_truth_pub = nh.advertise<nav_msgs::Odometry>(pub_topic, 1);
 
 	double utm_x, utm_y;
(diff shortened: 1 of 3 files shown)


igvc_gazebo/nodes/ground_truth/ground_truth_republisher.cpp
```diff
@@ -54,8 +77,9 @@ void groundTruthCallback(const nav_msgs::Odometry::ConstPtr& msg)
...
     g_ground_truth_pub.publish(result);
 
-    // publish transform for tf if there has not been a update from the localization node in the last second
-    // since it also publishes the same transform
+    // publish transform for tf if there has not been a update from the
+    // localization node in the last second since it also publishes the same
+    // transform
     if (std::abs(msg->header.stamp.toSec() - g_last_estimate.toSec()) > 1.0)
     {
       static tf::TransformBroadcaster br;
```
VAM7686, line 85: Maybe instead of having this as a static variable we could move this to the header file?

igvc_gazebo/nodes/ground_truth/ground_truth_republisher.cpp
```diff
@@ -67,7 +91,7 @@ void groundTruthCallback(const nav_msgs::Odometry::ConstPtr& msg)
   }
 }
 
-void utm_callback(const ros::TimerEvent& event, const tf::Transform& odom_to_utm)
+void GroundTruthRepublisher::utm_callback(const ros::TimerEvent &event, const tf::Transform &odom_to_utm)
 {
   static tf::TransformBroadcaster br;
   static tf::TransformListener tf_listener;
```
VAM7686, line 97: Also probably a good idea to keep these in the header file. Would probably need to rename the broadcasters to differentiate them and the respective transform that they broadcast.

igvc_gazebo/nodes/ground_truth/ground_truth_republisher.cpp
```diff
@@ -1,24 +1,47 @@
-#include <geometry_msgs/Vector3.h>
-#include <nav_msgs/Odometry.h>
+#include "ground_truth_republisher.h"
+#include <mutex>
```
VAM7686, line 2: Do we actually use mutex here?

igvc_gazebo/nodes/ground_truth/ground_truth_republisher.cpp
```diff
@@ -1,24 +1,47 @@
...
 #include <parameter_assertions/assertions.h>
-#include <robot_localization/navsat_conversions.h>
 #include <ros/ros.h>
```
VAM7686, line 4: These should probably be in the header file.

VAM7686 (changes requested):
Nice job! Just a few minor things that need to be addressed.

With regards to testing the functionality of the node I'd like to see a few unit tests. There's a ros node testing framework that can be found [here](https://github.com/RoboJackets/igvc-software/tree/master/igvc_platform/src/tests) and making a few tests should be pretty simple. A good example of some unit tests are the [differential drive unit tests](https://github.com/RoboJackets/igvc-software/blob/c0889878c1bd3f5b91b4511857b81aa3138e7eb8/igvc_platform/src/tests/test_differential_drive.cpp) which test the differential drive node.

PR #863, opened 2021-12-23
Added sensors to Swervi model

matthewhannay567:
# Description
Swervi model was missing key sesnors in the .xacro file, so I added them.

This PR does the following:
- Adds all sensors to Swervi model
- Changes the simulation launch files to spawn Swervi

# Testing steps (If relevant)
Run `roslaunch igvc_gazebo qual_low.launch`
Expectation: the sensors are visible on the sensor mast. (The lower IMU is inside the main body)

# Self Checklist

### DIFF
--- a/igvc_description/launch/spawn_swervi.launch
+++ b/igvc_description/launch/spawn_swervi.launch
@@ -6,12 +6,11 @@
     <arg name="roll" default="0.0"/>
     <arg name="pitch" default="0.0"/>
     <arg name="yaw" default="0.0"/>
+    <!-- accelerate lidar with GPU -->
+    <arg name="gpu" default="false"/>
 
     <!-- Model file -->
     <arg name="model" default="$(find igvc_description)/urdf/swervi.urdf.xacro"/>
-    
-    <!-- accelerate lidar with GPU -->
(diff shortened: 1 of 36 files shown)


igvc_description/launch/spawn_swervi.launch
```diff
@@ -15,7 +15,7 @@
     <!-- End of arguments list -->
 
     <!-- Spawn the vehicle. -->
-    <param name="robot_description" command="$(find xacro)/xacro $(arg model) gpu:=$(arg gpu)" />
```
VAM7686, line 18: Does this need to be changed? Isn't the default argument the same as the proposed change?

VAM7686 (changes requested):
All the sensors work. But, for some reason when I launch the simulation [the robot is shaking??](https://www.dropbox.com/s/eyffis97ze1nw5f/swervi_shaking.webm?dl=0)

igvc_description/launch/spawn_swervi.launch
VAM7686: actually this is fine

VAM7686 (approved):
Shaking issue should be solved by #867 

lgtm

PR #867, opened 2022-01-15
Swervi Wheel Shaking Bug

VAM7686:
# Description

Fixes the [wheel shaking bug](https://www.dropbox.com/s/eyffis97ze1nw5f/swervi_shaking.webm?dl=0) caused by the swervi gazebo PID controller by adding a simple filter to mitigate spikes in derivative value and help mitigate against sensor noise.

This PR does the following:
- Fixes wheel shaking bug by adding simple derivative filter.
- Prevents swervi gazebo controller dividing by zero.

# Testing steps
## Test Case 1
1. Run `roslaunch igvc_gazebo autonav.launch`
2. Zoom in on swerve modules.

Expectation: No evidence of shaking wheels.

## Test Case 2
1. Run `roslaunch igvc_gazebo autonav.launch`
2. Run `rostopic echo /front_left_effort_controller/command`

Expectation: No wildly oscillating effort commands on the `/front_left_effort_controller/command` topic.

# Self Checklist

### DIFF
--- a/igvc_gazebo/launch/swervi_control.launch
+++ b/igvc_gazebo/launch/swervi_control.launch
@@ -28,6 +28,7 @@
       <param name="wheel_radius" value="0.3429"/>
       <param name="max_effort" value="8.0"/>
       <param name="rate" value="60.0"/>
+      <param name="alpha" value="0.5"/>
   </node>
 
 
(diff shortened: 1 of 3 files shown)


matthewhannay567 (approved):
Works

Issue #492, opened 2019-09-11
Orientation in gazebo is off by 90 degrees


oswinso:
Our orientation when gazebo first launches is off by 90 degrees:
![orientation_off](https://user-images.githubusercontent.com/4125463/64665412-721c2580-d420-11e9-9fa8-82115c303f13.png)

PR #875, opened 2022-01-20
GPS Coordinates Fix

### LINKED ISSUE #874: Fix bug with hector gazebo GPS plugin
The hector gazebo GPS plugin is producing the wrong latitudes and longitudes in the simulation. Currently, along the +X axis (East) latitude increased and along the +Y axis (North) longitude decreased. Instead we need for +X to increase longitude and +Y to increase latitude.

[Here](https://www.dropbox.com/s/s5ar514tf0w5bt7/slam_incorrect.webm?dl=0) is a short video illustrating how the GPS coordinate (dark purple arrow) differs from the ground truth (green arrow). If the GPS coordinate is correct, both arrows should be in the same place.

VAM7686:
# Description

Fixes bug with GPS coordinates produced by the hector gazebo plugin. I think this bug was the root cause of #492 and #529 managed to patch it by adding an offset to the IMUs.

This PR does the following:
- Fixes bug with GPS coordinates
- Fixes all waypoint files whose coordinates were previously wrong due to the bug.
- Changes Swervi and Jessi IMU and Magnetometer yaw orientation to be 0 when facing north since most IMUs report 0 when facing north. I looked at the datasheet of the IMUs we currently have (Hillcrest Labs FSM-9) but couldn't find if this is actually true. Will look at this again during testing.
- Fixes `robot_localization` node by setting the correct yaw offset.

Fixes #874 

# Testing steps
## Test Case 1
1. Run `roslaunch igvc_gazebo qualification.launch track:=0` (tracks 0 through 10 available to test)
2. Run `roslaunch igvc_navigation navigation_simulation.launch`

Expectation: Robot navigates to correct locations in simulation as expected.

## Test Case 2
1. Run `roslaunch igvc_gazebo qualification.launch track:=0` (tracks 0 through 10 available to test)
2. Run `rostopic echo /fix`
3. Move robot in +X axis and observe longitude increase.
4. Move robot in +Y axis and observe latitude increase.

Expectation: Latitude and longitude increase as expected in respective axes.

# Self Checklist


matthewhannay567 (approved):
works

PR #884, opened 2022-04-10
Igvc gazebo/ground truth

### LINKED ISSUE #552: Add noise to simulation

dgorin6:
# Description

This feature adds noise to the x and y pose of the robot whenever groundTruthCallback() is called in the ground_truth node. The noise is set as a parameter in the simulation.launch file. The value of the noise represents the standard deviation of the normal distribution.

This PR does the following:
- Adds noise to x pose and y pose

Fixes #552

# Testing steps 
## Test Case 1
1. Set the value of x_noise_std_dev,y_noise_std_dev,z_noise_std_dev,roll_std_dev,pitch_std_dev, and yaw_std_dev  in simulation.launch to a non-zero value
2. rostopic echo/ground_truth and check position and RPY values

Expectation: Noise is added directly to the x and y pose

# Self Checklist


igvc_gazebo/nodes/ground_truth/main.cpp
```diff
@@ -36,8 +41,8 @@ void groundTruthCallback(const nav_msgs::Odometry::ConstPtr& msg)
     result.pose = msg->pose;
 
     // use the initial location as an offset (makes the starting location 0, 0)
-    result.pose.pose.position.x = msg->pose.pose.position.x - g_og_pose.pose.pose.position.x;
-    result.pose.pose.position.y = msg->pose.pose.position.y - g_og_pose.pose.pose.position.y;
+    result.pose.pose.position.x = msg->pose.pose.position.x - g_og_pose.pose.pose.position.x + x_distribution(engine);
+    result.pose.pose.position.y = msg->pose.pose.position.y - g_og_pose.pose.pose.position.y + y_distribution(engine);
```
VAM7686, line 45: To expand on this, you could try adding in more fields (z, roll, pitch, yaw) to experiment and see which fields is our stack the most sensitive to.

VAM7686 (changes requested):
Nice job! Just a few things that need to be addressed in addition to the comments:
- Make sure to add the link to this design document in the READMEs in `igvc-software/documents` and `igvc-software/design`.
- Add the new launch parameters for the ground truth republisher node in `simulation_low.launch` like you did for `simulation.launch`.

igvc_gazebo/nodes/ground_truth/main.cpp
```diff
@@ -25,28 +37,38 @@ void groundTruthCallback(const nav_msgs::Odometry::ConstPtr& msg)
...
 
-    tf::Quaternion quat;
+    tf::Quaternion quat(msg->pose.pose.orientation.x, msg->pose.pose.orientation.y, msg->pose.pose.orientation.z,
+                        msg->pose.pose.orientation.w);
+    tf::Matrix3x3 m(quat);
+    double roll, pitch, yaw;
+    m.getRPY(roll, pitch, yaw);
+    roll = roll + roll_distribution(engine);
+    pitch = pitch + pitch_distribution(engine);
+    yaw = yaw + yaw_distribution(engine);
```
VAM7686, line 69: Currently, this doesn't change the published message. You want something like below which constructs a new Quaternion using the RPY values plus noise.
```suggestion
    result.pose.pose.orientation = tf::createQuaternionMsgFromRollPitchYaw(
        roll + roll_distribution(engine), pitch + pitch_distribution(engine), yaw + yaw_distribution(engine));
```

VAM7686 (changes requested):
Looks good. Just a small fix is needed when adding noise to the orientation.

VAM7686 (approved):
Looks good to me!

PR #885, opened 2022-04-10
Remove Jessii remnants

matthewhannay567:
Removes / comments out things pertaining to Jessiii code, so that Swervi code is the default.


VAM7686 (approved):
lgtm

PR #886, opened 2022-05-03
3D Localization and Holonomic Teb Configuration

### LINKED ISSUE #662: Migrate to doing localization in 3D
#629 is going to require us to do localization in 3D (When we're on the ramp, the projections are going to be wrong if we assume we're always at `z=0` with pitch = 0).

While #555 is probably going to result in being able to perform 3D localization, to break a hard dependency with that issue I think we should look at the simpler method of using the current `robot_localization` ekf package to do localization in 3D.

The main things that need to be done will probably be:
- Turning the `2d_mode` parameter to `false` (we currently have it set to `true`)
- Checking that we're using the roll and pitch of the imu (check the "matrix" of bools)
- Tune some parameters or something / figure out what the issues are in terms of being able to get a "decent" 3D localization working.

AC: Working 3D localization in simulation.

### LINKED ISSUE #815: Change path planning parameters to work with Swervi
Because the change in drivetrain and size of the robot, we need to modify the parameters for our path planning to work with Swervi.

AC: Swervi is able to path plan without any problems and navigate to waypoints.

VAM7686:
# Description

Adds 3D localization using the `robot_localization` package as well as a new TEB configuration for swervi's holonomic motion

This PR does the following:
- 3D localization using the `robot_localization` package
- New TEB configuration for Swervi's holonomic motion
- Tuned PID parameters for the gazebo controller
- Adjusted height of Swervi so it can actually go over ramps instead of getting stuck on the hump <sup>lol</sup>

Fixes #815 and Fixes #662

# Testing steps
## Test Case 1
1. Run `roslaunch igvc_gazebo ramp_lane.launch`
2. Run `roslaunch igvc_navigation localization.launch`
3. Open RViz
4. Move Swervi to the peak of the ramp

Expectation: The /odometry/filtered pose should reflect the height and orientation of the robot on the ramp
image

## Test Case 2
1. Run `roslaunch igvc_gazebo autonav_low.launch`
2. Run `roslaunch igvc_navigation navigation_simulation.launch`

Expectation: Observe that Swervi is able to navigate while moving sideways (not only forward).

# Self Checklist


matthewhannay567 (changes requested):
I've verified that the localization works with height now, and the robot can move horizontally.
I also let it run in the simulation, and I managed to get over the ramp and past it. It got stuck in the part with the middle barrels however. Tune the parameters a bit more so it can get through.

matthewhannay567 (approved):
Going to go ahead and approve this. Robot moves well and managed to navigate most of the course. Fell off the second ramp though.

matthewhannay567:
Sorry for the delay. Ran it again, robot fell off the first ramp.

VAM7686:
I'm going to merge this in because holonomic teb and the 3D localization features work as intended. The reason why Swervi falls off the ramp is a problem with the traversability layer that deserves its own issue.

Essentially, the traversability layer understands the slope gradients when going up ramps but it does not understand the slopes as well when going down. When the robot is at a certain elevation it does not understand why going off a cliff is bad because the ground following the cliff has a gradient of zero (due to areas not seen by the lidar being classified as NaN cells) and is therefore traversable. Perhaps a way to solve this is using something like inpainting to fill in the NaN cells which will help define the boundaries more clearly. However, this approach classifies the top of the ramp as a barrier since the descending portion of the ramp is hidden by the ascending portion of the ramp so the mapping thinks that there is a cliff. Perhaps a modified idea of this can work.

PR #890, opened 2022-05-30
Drive encoders working on Swervi

AR2100:
Updated the .proto files for Swervi and updated motorcontroller.cpp such that /encoder topic is published. Tested on Swervi.


VAM7686 (approved):
Everything builds. Don't want to mess anything up.

